PR #11, opened 2017-12-03
Generate the prefix according to payload

jepsonrob:
You'll have to double check this works with your device as I haven't got the JSON data working for these smart bulbs yet, but it does make the prefixes work for me & allows for a response from the getStatus function.

This is my first pull request so please let me know if I've done anything wrong!


codetheweb:
Looks good, thanks.

PR #15, opened 2018-01-04
Doc Android root options for obtaining key

clach04:
Kudos to @nijave


codetheweb:
Excellent work, as always.  Thank you for helping keep this alive :).

PR #23, opened 2018-02-01
Add SSL certificate step

kaveet:
Some apps implement certificate pinning, which means they don't fully trust local CA certificates. As of iOS 10.3+ you might have to enable a toggle to get the app working with the Charles certificate.


clach04:
👍

codetheweb:
Thanks.

PR #26, opened 2018-02-10
Error handling of error (reasons)

unparagoned:


codetheweb:
Personally, I think this mucks up the basic usage documentation a bit too much.  If a function in the example does throw an error, it's pretty easy for a somewhat-experienced developer to track it down by adding `catch()` statements.

**TL;DR**: I like the idea but it makes the example too long.  If you have any ideas for a more concise version, I'm open to hearing them.

Issue #18, opened 2018-01-14
Infinite loop in TuyaDevice.prototype._extractJSON function


bobalob:
If the message contains more open braces "{" than close braces "}" the function goes into an infinite loop.

The real fix would be to clean up the decrypted message to remove the invalid brace at the end. I have seen similar characters at the end of messages when using zero padding in AES (not 100% sure this is the issue.) I cleaned up my messages by switching to PKCS7 padding.

A quick fix is to check for the number of close braces before entering the loop

```
    var leftBrackets = stringOccurrence(data, '{');
    var rightBrackets = stringOccurrence(data, '}');
    if (leftBrackets > rightBrackets) { leftBrackets = rightBrackets }
```

but this could end up with bad returned JSON if there happens to be other invalid '{' and '}' characters

codetheweb:
Yeah, I'm not completely sure what the best solution is here.  There's three possibilities:
- Formatting is perfect, none of the padding renders as an ASCII character and `numOfLeftBrackets == numOfRightBrackets`
- Random padding results in `numOfLeftBrackets > numOfRightBrackets`
- Random padding results in `numOfRightBrackets > numOfLeftBrackets`

I'm not really sure what to do.  One interesting thing to note is that I've never seen a `}` *before* the main payload, although that's most likely due to chance and not design.

The best way to parse returned data would be to simply start parsing at a certain index in the string like the actual device does, however this would add complexity as we would need to add the index to `requests.json`.

bobalob:
Is the message returned from the device encrypted or plain text?

codetheweb:
The majority of the time it is returned in plain text.  However, there seem to be a (very) few random times when it's returned encrypted.  I have no idea what triggers this.  If it's returned encrypted, a JSON error will be thrown.

codetheweb:
Fixed.

Issue #3, opened 2017-11-17
Unable to get parameters with Chales


lollo0296:
Hi together, 

I have a tuya smart plug and i am totally excited to integrate it in my homebridge configuration.

@codetheweb I have followed your SETUP.md but i can not get this POST request you are actually getting.

In my Charles after re-adding the device to my app i can see just this:
[https://imgur.com/a/F904M](url)
[https://imgur.com/a/jfJXa](url)

Can someone help me? 

Lorenzo.
(For reference you should invert point 4 and 5 of your guide)

codetheweb:
@lollo0296 did you setup the SSL certificate on your phone?  Without adding it, Charles cannot decrypt the encrypted traffic between your phone and device.

lollo0296:
@codetheweb thank you gor getting back so quickly.

Yes i did, you can take a look here https://imgur.com/a/Hm8Mz

Don’t know what is going on!
As I mentioned in the issue I wasn’t able to install the certificate before configuring the proxy on my iPhone, Charles website was having such an internal 404. 
I first configured the proxy, then I could install the certificate. That’s the reason for my suggestion to flip the order of point 4 and 5 on Setup.md.

lollo0296:
Up! No one knows what the problem could be?

@codetheweb no ideas?

codetheweb:
Sorry for not responding, I might have forgot about this 😛 .

Unfortunately, I don't really have any ideas.  I know you installed the certificate, but did you also [trust it](https://stackoverflow.com/a/43849537)?

lollo0296:
That can happen everyone 😂

No, I actually didn’t!! I will give it a shot and let you know :)

pierbiag:
FYI, apparently you now need to enable the proxying on each website for it to decrypt properly.
https://www.charlesproxy.com/documentation/faqs/ssl-changes/

codetheweb:
@pierbiag I'll hopefully update the docs soon to include this.  Did you get your plug working?

pierbiag:
I got the logger working but not the plug. In the end (without trying too much) I opened it up and reflashed the ESP8266 to use it with HomeAssistant.

codetheweb:
Nice.  It's good to know at least that these can be easily reflashed.

pierbiag:
[...]

PR #33, opened 2018-04-18
Improve packet parsing and avoid stalling

tjfontaine:
First commit: I was occasionally hitting an infinite loop in the JSON parsing (re #18),  I didn't spend time trying to debug that logic, instead I simplified the logic to merely strip off the prefix and suffix and attempt to JSON parse the remainder. Ideally we'd be looking for a field that indicates the payload size, which seems like what the 12th byte is, and it appears to scale with the payload but not exactly.

Second commit: while I have no evidence of this happening, it's certainly possible though that messages may be split over multiple `data` events, so I've changed the response handling logic to concatenate messages until there's a buffer that starts with the prefix and ends with the suffix.

What's still missing however, is the notion that there may be multiple messages in the stream, which neither `_extractJSON` or the parsing logic in `_send` are prepared to handle.

Third commit: Add a way to timeout waiting for a response from a device. In the event we have missed a message, it's possible for the state machine to stall (well there's no explicit state machine here, yet) so you'll end up with a potential socket leak, and unresolved promises.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,8 @@ const debug = require('debug')('TuyAPI');
 // Import requests for devices
 const requests = require('./requests.json');
 
+const parser = require('./message_parser');
+
 /**
 * Represents a Tuya device.
 * @class
@@ -97,7 +99,18 @@ TuyaDevice.prototype.resolveIds = function () {
     this.listener.on('message', message => {
       debug('Received UDP message.');
 
-      const thisId = this._extractJSON(message).gwId;
+      let p = new parser();
+      p.append(message);
+
+      if (!p.parse()) {
+        reject(new Error("Failed to parse message"));
+      }
+
+      let data = p.decode();
+
+      debug('UDP data', data);
+
+      const thisId = data.gwId;
 
       if (needIP.length > 0) {
         if (needIP.includes(thisId)) {
@@ -108,7 +121,7 @@ TuyaDevice.prototype.resolveIds = function () {
             return false;
           });
 
-          this.devices[deviceIndex].ip = this._extractJSON(message).ip;
+          this.devices[deviceIndex].ip = data.ip;
 
           needIP.splice(needIP.indexOf(thisId), 1);
         }
@@ -177,9 +190,6 @@ TuyaDevice.prototype.get = function (options) {
 
   return new Promise((resolve, reject) => {
     this._send(currentDevice.ip, buffer, this._responseTimeout).then(data => {
-      // Extract returned JSON
-      data = this._extractJSON(data);
-
       if (options !== undefined && options.schema === true) {
         resolve(data);
       } else {
@@ -313,20 +323,6 @@ TuyaDevice.prototype._send = function (ip, buffer, responseTimeout) {
           timeout = null;
         }, responseTimeout);
 
-        let buff = new Buffer(0);
-
-        function parse() {
-          if (buff.length > 8) {
-            let prefix = buff.readUInt32BE(0);
-            let suffix = buff.readUInt32BE(buff.length - 4);
-            if (prefix === 0x000055aa &&
-                suffix === 0x0000aa55) {
-              resolve(buff);
-              done();
-            }
-          }
-        }
-
         function error(e) {
           debug('failed to communicate', e);
           e.message = 'Error communicating with device. Make sure nothing else is trying to control it or connected to it.';
@@ -339,10 +335,16 @@ TuyaDevice.prototype._send = function (ip, buffer, responseTimeout) {
           client.destroy();
         }
 
+        let p = new parser();
+
         client.on('data', data => {
           debug('Received data back.');
-          buff = Buffer.concat([buff, data]);
-          parse();
+          p.append(data);
+          if (p.parse()) {
+            // TODO check for another message on the stream?
+            done();
+            resolve(p.decode());
+          }
         });
 
         client.on('error', err => {
@@ -368,27 +370,6 @@ TuyaDevice.prototype._constructBuffer = function (type, data, command) {
 
   // Create final buffer: prefix + data + suffix
   return Buffer.from(prefix + data.toString('hex') + requests[type].suffix, 'hex');
-};
-
-/**
-* Extracts JSON from a raw buffer and returns it as an object.
-* @private
-* @param {Buffer} data - buffer of data
-* @returns {Object} extracted object
-*/
-TuyaDevice.prototype._extractJSON = function (data) {
-  debug('Parsing this data to JSON: ', data.toString('hex'));
-
-  // every packet is wrapped by 20 bytes including the prefix
-  // and 8 bytes including the suffix, and there should be at least
-  // 2 more bytes for { and } -- 30 = 20 + 8 + 2
-  if (data.length < 30) {
-    throw new Error("malformed data packet");
-  }
-
-  // valid JSON should be in the inner payload
-  data = data.slice(20, data.length - 8);
-  return JSON.parse(data.toString());
 };
 
 module.exports = TuyaDevice;
--- a/index.js
+++ b/index.js
@@ -66,6 +66,8 @@ function TuyaDevice(options) {
   this._connectTotalTimeout = undefined;
   this._connectRetryAttempts = undefined;
 
+  this._responseTimeout = 10 * 1000;
+
   debug('Device(s): ');
   debug(this.devices);
 }
@@ -174,7 +176,7 @@ TuyaDevice.prototype.get = function (options) {
   const buffer = this._constructBuffer(currentDevice.type, thisData, 'status');
 
   return new Promise((resolve, reject) => {
-    this._send(currentDevice.ip, buffer).then(data => {
+    this._send(currentDevice.ip, buffer, this._responseTimeout).then(data => {
       // Extract returned JSON
       data = this._extractJSON(data);
 
@@ -265,7 +267,7 @@ TuyaDevice.prototype.set = function (options) {
 
   // Send request to change status
   return new Promise((resolve, reject) => {
-    this._send(currentDevice.ip, buffer).then(() => {
+    this._send(currentDevice.ip, buffer, this._responseTimeout).then(() => {
       resolve(true);
     }).catch(err => {
       reject(err);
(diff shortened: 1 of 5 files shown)


tjfontaine:
Correction, 16th byte or 4th word appears to be how much to read for the rest of the message, and the word before the suffix is the crc32 of the payload

tjfontaine:
Mostly for feedback purposes I added a fourth commit, that does the CRC checking and factors the parsing out into its own file.

One could also use this for message encoding and not just decoding.

codetheweb:
Wow, thanks for your contributions, this is a lot to take in. 😃 

After skimming your changes, it appears you're extracting data based on hard coded values?  While I'm not against that, I'm unsure if that's gonna be the same for every device.  Maybe (at least until we get conformation that this works for multiple device types) we could have two "parser" options: the current one that grabs data between `{` - `}`, and your parser that looks for prefixes and suffixes, with the later being the default.

What do you mean by "there may be multiple messages in the stream"?

Thanks for all the little fixes.

codetheweb:
Ok, got it.

I tested your fork locally, and I'm having some issues.  Apparently my device always returns
`
00 00 55 aa 00 00 00 00 00 00 00 07 00 00 00 0c 00 00 00 00 78 93 70 91 00 00 aa 55
`
after a set command is sent, no matter the current status.  Is this different from your device?  Currently a JSON error is being thrown as there's nothing to parse.

tjfontaine:
Right, so `decode` shouldn't fail when there is no body, if we look at the structure:

```hex
000055aa
00000000
00000007
0000000c
00000000
78937091
0000aa55
```

We were expecting 12 byes, which is just enough for the (presumed) null padding before the payload, the CRC, and the suffix/trailer.

So 0x7 is probably an `ACK` saying the device received the message?

tjfontaine:
I've added a commit that includes your `ACK` packet fix

BillSobel:
@tjfontaine  While I agree that the integers are large given the payload, the sizes and layout comes from disassembly of the source.  TuyaFrame in the Android app specifically (and from that what I called version is indeed most likely a int sequence number).  There are a number of 'strange' decisions made in the implementations (such as which encoding to the cloud API's to use is based not on the API called, but rather on what kind of key you have [cloud key, hardware key, app key])...

Also, from my review, there is the possibility of more than one response packets on the connection.  If a status command is sent to the device and the TCP connection is left open, the beacon packets that was previously broadcast are instead returned every few seconds on the TCP connection.  This continues until the connection is closed and then the device reverts to UDP broadcast.

BillSobel:
@tjfontaine 0x07 I believe is a status response (e.g. state reporting), this is the stream of data you will get if the connection is left open.  The mcu data is wrapped in a packet.  The mcu definition:

"3.8 state Reporting

Notes：

1.See “3.7” for datapoint state data units.

2.“state reporting” is an asynchronous treatment protocol and there are three trigger mechanisms for “state reporting”:

​	When the MCU receives the “Command issuing handle frame”, it will execute correctly corresponding datapoint command and then “state reporting”, finally the frame will send the changed datapoint state to the module

​	After the MCU detects any change of datapoint, it will send the changed datapoint state to the module

​	When the MCU receives the state query frame in 3.9, all datapoint state will be sent to the module. 

3.“Command reporting” can contain many “state data units” of datapoint"

See https://docs.tuya.com/en/mcu/mcu-protocol.html

tjfontaine:
@BillSobel ah thanks, I was just observing what I could see from the packets that this package was already capturing, I hadn't sought out any documentation or done any disassembly. However the documentation requires sifting through some ambiguity, I'll have to assume most of this is a factor of the translation.

Given this, I think I can come up with a better way to handle command encoding/decoding, though it may make sense to do a different PR

codetheweb:
@BillSobel thanks for that link, I clicked around on their documentation site and hit the jackpot.  [This file](https://images.tuyacn.com/smart/docs/1490928897-tysdk_for_esp8266_1.0.5.tar.gz) is the source of the firmware loaded on their ESP8266s.  I've just started skimming over it, but so far it looks like it's based on / loaded in [FreeRTOS](https://freertos.org/).  All the Tuya-specific files seem to be in `app/`.

@tjfontaine If you think you can use the data length field in the packet for parsing, I'll close this pull request.

BillSobel:
@tjfontaine My pleasure, glad to not be the only one tilting at this windmill ,) I did get official cloud and client keys from Tuya, but am debating using them vs the Android app key (Im writing a plugin for Homeseer and can either be fully standalone or compatible with the Android app).  The documentation is poor, typos, api names mis-transposed.  The definition of the different algorithms is a mess (but as I mentioned, my personal breakthrough was realizing the signing format was based on which key you used not which API you called).  I did get wifi registration working (Access point mode).  Do you (or anyone) have any idea how the other mode (smartlink) works?  Bluetooth?  Or some custom configuration API that android and iOS support for wifi devices?

@codetheweb Thanks, I had missed the firmware image.  That should be very useful.

codetheweb:
@BillSobel what do you mean you got WiFi registration working?  Are you able to do it entirely locally, or do you have to communicate with Tuya's API to get the device fully registered?

BillSobel:
@codetheweb  Registration requires communicating with the Tuya API.  So what I have done is getting the token from Tuya, sending it to the device along with wifi info (ssid/password).  The device immediately then connects to Tuya (if the credentials are good) and I can then get the local key from the Tuya API by querying on the same token I sent to the device.

codetheweb:
@BillSobel ok.  From your current understanding, is there a way to set WiFi parameters on a device and generate/set a valid localKey without ever touching the cloud API?  Ideally there would be an easy way for users to get this working besides capturing packets or getting cloud API credentials.

BillSobel:
@codetheweb Unfortunately without a firmware change, absolutely not.  The cloud has to participate in  registration to get the local key.  That leaves us with the key problem.  I've spent a few (frustrating) days on this . I originally thought I had the Android app keys, but I don't (explained below).  

First, if it is not clear:  Every group of keys puts devices into their own silo in the Tuya (AWS) backend.  Meaning if I use my one of my keys to register a device, that device is only visible to tools/apps that use one of the other keys in my group (e.g. my android key is different than my iOS key, but it appears that registering a device on Android still makes it available to iOS).  But importantly, it is NOT visible to other key groups.  So my devices registered on my Android key are NOT visible in the Tuya application.  In fact I have completely different logins (using the same email but different passwords) under the SDK keys (see below) and my assigned keys.  So this is a big issue if folks want to use any Tuya provided tools to control their devices or different ventures like this use different keys, we will all be locked into the tool we register with (if we can't assign keys to the tool ourselves)

Here are our realistic access key/secret options as I see them:

1) Everyone applies to Tuya individually.  They do want project information and I am unclear how many individuals they would approve (I was approved and have my 'own' keys).  Users need to supply keys to any tools they want to use.  No interop with official applications. No ethical issues. 
2) We jointly apply for a project or open source key for these types of projects to use.  Key can be included in the tools. No interop with official applications. No ethical issues.
3) We use the SDK keys.  This wasn't obvious, but Tuya has posted keys for use.  The android SDK is available for you to download after simply signing up with Tuya and it does have a getAppSecret and getAppKey which return a valid key (but this key is NOT the same as the Tuya apps use, so all registrations would go into the SDK silo).  No interop with official applications.. Light grey ethical issues (but we could ask Tuya for verification this is the intended purpose of the keys, it certainly seems that way to me.  They are NOT protected like the Tuya apk keys are.)  Keys can be included in the tools, and the eco system of hobbies are all using the same keys (so tools can be shared and will work with each persons registrations.  
4) We extract the keys from the Android or iOS apps.  I looked into this, as it allows compatibility with the Tuya apps for people who want them.  However (I've only looked at Android) Tuya has gone through great lengths to hide the keys.  Someone posed a secret (46) and a UUID that went with it.  Those aren't the keys, those are inputs into a security library from Alibaba (see https://jaq-doc.alibaba.com/docs/doc.htm?spm=a313e.7629140.0.0.hTjhEi&treeId=243&articleId=105581&docType=1) The library does the signing internally in native code using lookup tables, the key itself is ended into a png image that is shipped as application resources.  The library never releases the key outside of the native code.  Worse (and this just annoys the heck out of me)  Tuya is using a completely different signing mechanism for their Application than has been documented on the website for the other keys.  The input strings sent to be signed are padded with some unknown data (the tail padding starts with az so I think its a token provided by the users region)  Anyhow, extracting the keys went from a few hours to probably a few weeks (albeit, I was pretty close to just running a signing service on the hooked apis since i have the android app running under emulation anyhow ,))  Full compatibility with Tuya applications.  Morally dark grey (I won't say black because the end goal of all this work sells more Tuya devices and you can only touch your own data if you have your login credentials with the keys in question, so even if we have them I can't go and attack you and vice versa).  Keys could be included in the tools and be fully compatible with the official applications.

BillSobel:
@codetheweb I should add, for whatever its worth, my personal thought is key option #3

codetheweb:
Thanks for the background information @BillSobel, I assumed that if someone created a device it would be available on a different app under the same login.

As I've [stated before](https://github.com/codetheweb/tuyapi/issues/5#issuecomment-349685543), I don't want to extract production keys from the app (and it doesn't sound like Tuya wants you to either 😛).  However, the demo key option sounds interesting.  I say we use it, and if Tuya complains then we can do something different.

Also: if we are able to make something for configuring WiFi and getting a localKey, I would prefer that be done in a different codebase (and probably with a CLI interface for now).  I'm thinking about creating a Github TuyAPI organization (instead of it all being under my username) to help make it clear that I'm not the only one working on this and to make it easier for other people to contribute (you don't have to open pull requests once you're approved, just push to the repo).

codetheweb:
I don't think the firmware "source" is as useful as I thought it was originally.  Looking through it, it's all header files.  I'm not that familiar with C++ libraries, so I'm not sure if this is common practice or how it works, but I can't find any actual source code in the image, just function definitions.

codetheweb:
@BillSobel do you happen to have code for the WiFi registration step you got working?  I don't care how messy it is 😛.

BillSobel:
@codetheweb I do, but the code is not from an OS project.  So I'm limited on sharing code, but am free to share the details.  Where did you get stuck?

codetheweb:
I'm gonna go ahead and merge this.

@tjfontaine if you're still working on this, let me know.  Otherwise I'll look into modifying parsing based on the payload size field.

PR #36, opened 2018-05-05
Optimize "resolveIds"

vkoop:
If for all devices IPs are set the method "resolveIds" can return true immediately without opening a socket.


codetheweb:
👍

Should have realized this myself.

PR #88, opened 2018-10-12
Add persistent connection with event emitting

Apollon77:
- enhanced library to support persistent connection to leave connection open and keep them alive with ping/pong messages
- with persistentConnection data and other events will be emitted as events
- Events are "data", "connected", "disconnected" and "error"
- When using persistentConnection get method only return true on success because data are emitted as event
- When not using persistentConnection everything works as before
- enhanced message parser to also return the commandByte from the response to allow checking/to react on it
- When using persistentConnection an automatic reconnect happens when disconnecting

Based on the great work of @NorthernMan54


Apollon77:
I tested it with this script additionally to the default example:

```
const TuyaDevice = require('./index.js');

let tuya = new TuyaDevice({
  id: 'xxxx',
  key: 'xxx',
  ip: 'xxxx',
  persistentConnection: true
});

let cnt = 0;
let status;
tuya.on('data',(data) => {
  console.log('EVENT DATA ' + cnt + ': ' + JSON.stringify(data));

  if (data && data.dps && data.dps["1"] !== undefined) status = data.dps["1"];

  if (cnt++%30 === 0 && status !== undefined) {
    console.log('Current Status: ' + status);
    tuya.set({set: !status}).then(result => {
      console.log('Result of setting status to ' + !status + ': ' + result);
    }).catch(err => {
      console.log(err.message);
    });
  }

});

tuya.on('connected',() => {
  console.log('EVENT CONNECTED');
});

tuya.on('disconnected',() => {
  console.log('EVENT DISCONNECTED');
});

tuya.on('error',(err) => {
  console.log('EVENT ERROR: ' + err);
});

// setTimeout(() => {tuya.disconnect();}, 30000);

tuya.connect(); // get or connect to start

```

Apollon77:
Sorry for all those additional codestyle fix commits ... I did not had it on radar that it was configured that strict :-) but now everything is green again

NorthernMan54:
If you give me a few days I can run this thru my test environment and see how it works.  Earliest is likely Sunday.

Apollon77:
Sure. Any feedback is welcome

codetheweb:
Hey @NorthernMan54 and @Apollon77, thanks for all the work.  I just glanced over it (I won't be able to fully test it for a few days), but noticed a few things:
1. @Apollon77, you can run the linting tests and the code tests locally with `npm test`.
2. `npm run document` should be ran whenever the inline-documentation is updated so the docs pages will update.
3. It looks like the command byte option for Parser was changed from a hexadecimal string to a integer that's then converted.  Why is this?

I'll try to look at this more in depth soon, I really want to merge it. :)

Apollon77:
1+2 i will check later. Also if docs make sense for the new methods. 

3: in fact it is a byte and you just implemented it as string to be added to the buffer as string before converting to binary buffer. Because I also wanted to return the result commandbyte to compare and also expose in the data event it seemed strange to returns a zero padded string there ;-) a number (as it is) feeler more natural ;-) and so i just changed it because it was not much. Makes sense?

I had it run with the above code all over the night. Also reconnection worked great.

Apollon77:
@codetheweb are there best practices how to document the emitted events? They are not functions...

codetheweb:
@Apollon77 take a look at these pages: [@fires](http://usejsdoc.org/tags-fires.html) and [@event](http://usejsdoc.org/tags-event.html).

Apollon77:
will add tomorrow

Apollon77:
Ok, docs updated too ... please check.

Apollon77:
to 1.) removed, I also found this page in web but it seems that this is more the "internal" protocol because the commands do not match
to 2.) I was also struggling with this. a command X without data is kind of the "Ack" of the message before. But you are right, there is only limited benefit outside of the library, so I do not emit empty data now.
to 3.) I also had that once in a while. It seems that the dev ice sometimes closes the connection when messages coming too fast or because of other reasons - and sometimes then it refuses connections (maybe it restarts internally?), but this fixes itself after some time. Please check the library, but my experience so far was that it was always caused by the device itself.

NorthernMan54:
@Apollon77 @codetheweb I think we can publish this.  I ran the latest version tonight for about 6 hours, and it worked perfectly.  And didn't drop the connection at all, and reconnected properly when resetting the device.  I think yesterday my home network was having something issues.

codetheweb:
Looks good, guys.  Sorry I haven't been able to test this yet but I've been pretty busy.

This (baring any major issues) will be merged by the end of the weekend.

codetheweb:
Just pulled this and really liking it so far.  It's pretty slick that it emits events if you physically change the state. :)

A few things:
- If I unplug the device while `tuyapi` is connected to it, both error and disconnected events are emitted.  It should be one or the other (probably an error event).  Not sure what would be the best way to implement this.
- `index.js` is getting really long; do you guys have any ideas on how to further split it up?  (I know the inline docs are a bit of a pain as well and are one of the main contributors to the length issue, so if you have any input on that I'm open to hearing it.)
- There's some issues with the style and docs, but those can be fixed later (probably by myself).

NorthernMan54:
@Apollon77 regarding number 1, in my original I only didn't implement disconnected, and just used error for all issues with client, what's your thoughts?

Apollon77:
to 1.) I see both events in general unconnected! When the client closes the connection in a clean way then there only is a disconnected event and no error. In theroy the error event is also generic and could be used to report other type of errors as well to the caller like "can not parse message" or stuff. I have no idea if an "error" event "on network socket can happen without a disconnect" :-)

It is also a quesdtion what you do with these events. A "disconnected" event, when consumed, will maybe start different logic then a "error" event. I use "error" events mostly to log stuff in a meaningful way, but on disconnect I start a timer to see if I get a connected soon and if not have recovery logic or stuff.
So the events have a different purpose and give you different triggers to react on for special purposes. mixing this feels weired.

 I here oriented also on other libs where this is unconnected and that why I designed it that way. Makes sense for you?

to 2.) I also thought about that. The only way I could think of was to move the "connect/disconnect" layer to a separate class to be used internally and wrapped by the real API, but there is no real benefit beside "file gets smaller" and it would make things way more complex because you need to "proxy" aso the events and such ... so I decided not to split it up

to 3.) Tell me what (tests are green, so fullfills the automated tested part :-) ) and I can check it ... else I'm also happe if you fix it directly :-)

codetheweb:
Currently trying to run through this so it's merged tonight.

What do you guys think about making `persistentConnection` default to `true` instead of `false`?

Apollon77:
I did it with false because of backward compatibility... users need to change their code. 
From my point of view possible when you change major version ;-)

codetheweb:
Ok.  I do want to retain backwards compatibility, but I want to somehow encourage people to use the new connection scheme.

Maybe updating the example in the README will be enough.

Also, `get()` behavior seems to be broken when using a persistent connection.  Why did you make it always return true?

Apollon77:
Idea was to have get just trigger the request ... response will come as event then. That’s why „true“ ... and exactly here users need to change their code. 

Alternative would be to guess the response to a request ... not 100% accurate in worst case.

Apollon77:
Best would be two examples ... the Event one first ;-)

codetheweb:
Ok... I get what you're saying, but a purely event-driven model breaks down as code complexity grows.  For example, what if you have two different bits of logic that control the same device?  It doesn't really make sense to have a single event handler for both.

It is convenient in some situations, so I'd say keep the data event but make the `set()` and `get()` functions properly return on event emit.

Ideally, the code below would work (draft for example in README):
```javascript
const TuyaDevice = require('tuyapi');

// Make new instance of class
let device = new TuyaDevice({
  id: 'xxxxxxxxxxxxxxxxxxxx',
  key: 'xxxxxxxxxxxxxxxx',
  ip: 'xxx.xxx.xxx.xxx'});

// Initialize socket connection to device
device.connect();

device.on('connected', () => {
  console.log('Connected to device.');

  device.get().then(status => {
    console.log('Current status is', status);

    device.set({set: !status}).then(result => {
      console.log('Result of setting status', result);

      // Clean up the socket to exit gracefully
      device.disconnect();
    });
  });
});

```

Apollon77:
So, I added a commit to my branch ( dont know why it is not showing up here?!)
https://github.com/Apollon77/tuyapi/commit/e1d52e882980f20be187f6228d21d172be835ca9

This changes the way messages are handled to be like requested:
* get will return data as before in promise - unless options.returnAsEvent=true is set (mainly used internally to return first get  as event)
* set will return true as before, BUT will do that after the ack from the device was received. So no "fire and forget" set as before, but a real one :-)

This means in fact that most of the emitted "data" are commandByte 8 because actively pushed by the device.

I hope the commit also shows up here soon ... else I close this PR and create a new one?

Apollon77:
But: To note (and maybe also something for documentation?) Users should be aware that only one command to the device after each other is supported for now. So if someone manages to "Mix" calls (mainly get/set) to run "parallel" then he may not receive a promise resolve/reject for one of them.
Option would only be to implement a queue or such, but maybe it's not worth the effort

NorthernMan54:
Max,

This is in response to your comments around using an event driven programming model.  As the power of event driven is extremely powerful, it requires the developer to get into the right mindset to work with it, as it is dramatically different then procedural model.

When I think about event driven programming I use the analogy of the mail man delivering mail to your mail box, and you wanting to know when it has arrived.

Without utilizing an event model, you need to sit and watch/wait for the mail man to put something in your mailbox, and during this time that your waiting you can’t be do anything else.  ( including playing with your phone )

In an event driven model, the mail man rings the doorbell, at the same time as he drops the mail into the mailbox.  So if you were doing something else, you can drop it and go get the mail.  Which is more efficient.

Where the power of events really comes to bear, is when you need to watch multiple mail boxes, which without an event based model gets really complicated with polling loops etc.

In a very simple interaction model, with a single device where it doesn’t generate its own events a get/set model works exceptionally well.  But once you get into devices that have the ability to generate local events it quickly turns into a mess of device polling loops etc.

With all this said, there is value in both methods are they are very different scenarios and interaction models they are dealing with.

For me, as my Tuya devices are replacement wall switches, with a local control switch, I need the event model so that homebridge/HomeKit always has the current state.  If your Tuya devices don’t have a local control option, events only add complexity with no value, and the get/set pattern is best.

In my homebridge plugins, I’m religiously avoiding using get methods, and only using events.  As get methods introduce slowdowns into HomeKit response times.  Homebridge will cache the current value/state of a plugin, so no need for a get method.

Sorry for the wall of text.

Apollon77:
Great explanation and analogy ;-)

The new version should also work good er also better then before. Correct is zwar you basically do not need get at all and we should avoid in the example and use events there. 

An implicit get is called on connection establish and return the current values as event. Then you get local updates as events and also after calling set to change the value you get an event that additionally brings you the real updates values.

Apollon77:
@NorthernMan54 Can you give the new version a try ?!

codetheweb:
Thanks guys for the feedback.

Thank you @NorthernMan54 for the great analogy, it's not that I am against event-driven programming but as you said it's beneficial to include both options.  For example, many types of devices other than basic sockets, such as a plug with power metering, need an implicit `get()` function because they don't emit events to connected sockets on power draw change.

I have some open time later today, I'll take a look at your changes @Apollon77.

Apollon77:
The new code will not create event on get but return as promise. (Old Version did Return result only as event, but you were right it is kind of unexpected). Personally in this case I think the new way is the better solution.

codetheweb:
Hey @Apollon77 I'm trying to push some changes to your `events` branch, but apparently because your repo is forked from NorthernMan54's the permissions system is messed up and I'm getting a permission denied error.  Can you please [add me to your repo](https://help.github.com/articles/inviting-collaborators-to-a-personal-repository/)?

codetheweb:
Thanks.

Apollon77:
I added some comments to your commit ...

codetheweb:
I replied @Apollon77, not sure if you got a notification.

Apollon77:
@codetheweb me too and yes I was wrong. Your docs make 100% sense :-)

codetheweb:
How does the [updated README](https://github.com/Apollon77/tuyapi/tree/events) look @Apollon77?

Apollon77:
Will check later tonight when kids are in bed ... now family time :-)

Apollon77:
README is ok. Other idea could be to issue the "set" in "on connected" and do not do a disconnect at all or with a timeout after 60 seconds or so ...

codetheweb:
The issue with that @Apollon77 is that the connected event may be emitted multiple times within the same session (at least it is for my device, likely a result of my school's weird network setup).

Apollon77:
Sie but it is only the example and to disconnect after the first event do not really show the power of the event stuff. 
You added the info about connect and this is fine in my eyes.

NorthernMan54:
I took a look at the readme and it looks great, my only comment was in the event example we had the set inside the event, when it didn’t need to be.  Should we move it outside?

Apollon77:
See my idea: put in connected ;-)

codetheweb:
What do you mean @NorthernMan54?  If the example is supposed to change the first property of a device, the `set()` needs to be inside the `data` event, no?

@Apollon77 I would be fine doing that, but (at least for my setup) doing so results in my device clicking on and off intermittently (as the connection is lost and re-established) until I exit the script.  I feel like if that happens for other people it might be confusing and put them off using events.

Apollon77:
Ok then place the set outside of everything once with a timeout of some secs maybe?!

codetheweb:
@Apollon77 @NorthernMan54 do you mean something like this:

```javascript

const device = new TuyAPI({
  id: 'xxxxxxxxxxxxxxxxxxxx',
  key: 'xxxxxxxxxxxxxxxx',
  ip: 'xxx.xxx.xxx.xxx',
  persistentConnection: true});

device.on('connected',() => {
  console.log('Connected to device.');
});

device.on('disconnected',() => {
  console.log('Disconnected from device.');
});

device.on('data', data => {
  console.log('Data from device:', data);

  const status = data.dps['1'];

  console.log('Current status:', status);
});

device.set({set: false}).then(result => {
  console.log('Result of setting status:', result);
});

device.on('error',(err) => {
  console.log('Error: ' + err);
});

device.connect();

setTimeout(() => { tuya.disconnect(); }, 30000);

```
?

The issue with that is that putting the set outside of the `#data` event means that the script can't dynamically set a property on the device.  Maybe you could write an example with what you're thinking?

Apollon77:
I mean it That way. This example would work for me because it shows event stuff and that you still can use set method. 
From my point of view:go ;-)

codetheweb:
Ok.  In the interest of trying to keep it as similar in effect as the other example, I think I'll leave the `set()` inside the event for now, although I will add the `setTimeout`.

codetheweb:
Gonna go ahead and merge this now.  Thanks again for all the great work @NorthernMan54 and @Apollon77.

(Side note: the main file is now over 500 lines long.  I sense a need for some refactoring if either of you are interested in helping. 😉)

codetheweb:
New version now published on NPM: v3.1.0.

PR #97, opened 2018-10-28
Update Travis CI to test for older language version

codetheweb:


Apollon77:
I would do 
- 6
- 8
- 10

PR #100, opened 2018-10-30
Update SETUP.md

neojski:
I just verified it all worked fine on my Samsung S9.


codetheweb:
Huh, great.  I read online that it wasn't possible to trust root certificates on Android devices which is why I added that. 🤷‍♂️ 

Thanks for testing.

PR #102, opened 2018-11-02
catch errors from parser and emit error if needed

### LINKED ISSUE #101: Exceptions thrown by messageparser are not catch by main class
The messageparser class throws an error in parse if magic suffix or prefix do not match. This is ok, BUT the index.js need to catch the exception and emit an error event instead because else the exception is unhandled (and also not really catchable from outside) and so the whole application it is used crashs

Apollon77:
fixes #101

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -2,7 +2,7 @@
 <html>
 <head>
   <meta charset='utf-8' />
-  <title>tuyapi 3.1.0 | Documentation</title>
+  <title>tuyapi 3.1.1 | Documentation</title>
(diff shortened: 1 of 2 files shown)


Apollon77:
I prepared the PR in Github web only because do not have my working environment today. So please regenerate the doscs :-) thank you.

A fast release would be great. Thank you

codetheweb:
👍 thanks.

Apollon77:
Thank you again for an info on npm publish :-)

codetheweb:
Just published @3.1.2. 😉

PR #105, opened 2018-11-06
handle error nicely

jpillora:
at this point, when i provide a bad gw id, my device returns the string `gw id invalid` here and therefore, a crash:

```
TypeError: Cannot read property '1' of undefined
    at _send.then.data (/.../Code/Node/tuyapi/node_modules/tuyapi/index.js:181:27)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```


codetheweb:
👍 thanks.

PR #116, opened 2018-12-01
Queue calls to resolveIds

### LINKED ISSUE #114: IP discovery binds to static port
I used this API to control my Tuya sockets with a Raspberry Pi and was delighted to find that the control is instantaneous—as fast as a light switch—but I can't seem to do this for more than one light. Your doc states that we can't have more than one TCP connection per device but is there a way to maintain persistent connections to multiple devices? If not, we must connect each time, and the result is that it takes ~3-4 secs to turn a light off. How does the Tuya app do it?

Thanks!

neojski:
Fixes #114.

My test case was roughly like this:

```
function socket ({id, key}) {
  const device = new TuyAPI({id, key, persistentConnection: true});
  device.resolveId().then(() => {
    console.log('success');
  }).catch((error) => {
    console.error(error);
  });
}
const tvSocket = socket({id: 'device1', key: '...'});
const downHeatingSocket = socket({id: 'device2', key: '...'});
```

It consistently failed with the old code and consistently succeeds with the new code.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,67 @@ const debug = require('debug')('TuyAPI');
 const Cipher = require('./lib/cipher');
 const Parser = require('./lib/message-parser');
 
+function resolveId (device, options) {
+  // Create new listener
+  let listener = dgram.createSocket('udp4');
+  listener.bind(6666);
+
+  debug('Finding IP for device ' + device.id);
+
+  // Find IP for device
+  return timeout(new Promise((resolve, reject) => { // Timeout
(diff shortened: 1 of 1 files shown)


codetheweb:
👍 looks good, thank you.

Wasn't able to test it myself as I don't currently have access to more than one device, but it looks like it should work fine.

codetheweb:
Published in `v3.1.5`.

neojski:
Thank you!

PR #140, opened 2019-02-08
Added functionality to resolve id if only ip provided. Also checks if id/ip are empty strings

unparagoned:
I noticed if an empty string is used to initialise id/ip then the check if undefined gets bypassed. 
There were two parts
1. I added functionality to resolveID() so that it will resolve either the id or ip if the other is missing.
2. Changed some checks
*I added a function to check if it's undefined, is it a string and if it's length is greater than 0.
*I changed some of the initialisation to use the new check function.
*I  removed some throw errors since a missing id can now be resolved. Also all get commands work without a key.


unparagoned:
Errors

PR #141, opened 2019-02-08
ResolveId will resolve id/ip from the other. Added and removed some checking.

unparagoned:
I noticed if an empty string is used to initialise id/ip then the check if undefined gets bypassed.
There were two parts

1. I added functionality to resolveID() so that it will resolve either the id or ip if the other is missing.
2. Changed some checks
   *I added a function to check if it's undefined, is it a string and if it's length is greater than 0.
   *I changed some of the initialisation to use the new check function.
   *I  removed some throw errors since a missing id can now be resolved. Also all get commands work without a key.

I don't really know the process of pull requests and looks like I've made some mistakes.


codetheweb:
Thanks for your work @unparagoned.

I'm actually in the process of migrating TuyAPI to TypeScript, which should hopefully remove the need for most of these checks.

I may still merge this if TypeScript ends up not working out for some reason.

codetheweb:
After converting the TuyaCipher class to TypeScript, I've realized that while it does offer some benefits, they won't ever be felt by consumers of this library (i.e. compiled TypeScript does not have automatic type checks like I thought it did).

I think for now the costs of moving to TypeScript outweigh the benefits, so I'll go ahead and merge this.

I'll also take a closer look at the get/set functions.

(Note: not sure if you figured this out, but you can run the tests locally with `npm test` 😉.)

unparagoned:
I saw the npm run cover after my initial version. Also is there a way to to get my linter to use those rules, since there were some differences even when I got that working. What I had issue with was trying to get test.js running. Also wasn't sure if there were more detailed unit tests I could use to check to make sure any changes I made haven't broken anything. As in tests to get the state of my device, set the state, using normal and dps settings, etc.

codetheweb:
There may be a XO integration for your IDE depending on what you're using, but what I usually do is just run `xo --fix` which automatically takes care of most formatting errors.

As long as you ran `npm i`, tests should get run with Ava when you do `npm test`.  Let me know if that doesn't work.

As for more extensive tests, another user has been working on a module called tuya-mock that implements a stub device for local testing.  I forked it [here](https://github.com/codetheweb/tuya-mock) and have been working on it.  However, there is a bit of an issue as described [here](https://github.com/dustinmcbride/tuya-mock/issues/1).  Please contribute if you have any thoughts.

PR #147, opened 2019-02-17
Fixed example code. Added device.connect

unparagoned:
Since you removed connect being called automatically, you now have to manually call connect. So You can update the read me or change it back. Either way using the synchronous code examples will have two gets called in close succession. I don't know if it's a real issue, I do get some read  errors now and then but they could be for any reason. I've change my code to be more asynchronous, so I don't have these issues anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ const device = new TuyAPI({
 (async () => {
   await device.find();
 
+  await device.connect();
+
   let status = await device.get();
 
   console.log(`Current status: ${status}.`);


codetheweb:
Oops, good point.  Thank you.

You may be right, I've been getting a few connection errors with v4 and I'm not sure where they're coming from.  I'll have to do further testing.

unparagoned:
@codetheweb Initially in v3 get called connect and connect called get. It all depended on how it was set up, etc. So I think I've figured out how to get it to do what you want in all circumstances. The behaviour you want is actually for connect and get to do the same thing. Connect if required and then get the state. The best way of getting it to work consistently and reliably is to rename connect to directConnect. Have connect just calling get. Have get calling directConnect. So when you run either connect or get, it tried directConnect(if it's already open it does nothing) otherwise it's directConnects to device, then gets the state.

codetheweb:
That's pretty much what it should be doing right now, with the only difference being that `.get()` no longer calls `.connect()` (which I think is good behavior).

I was commenting more on the fact that if the first connect attempt fails, it appears TuyAPI may not connect correctly on a second or third attempt when the device is actually ready.

unparagoned:
Oh, I thought you were talking about connecting to the port. I haven't had any issues with connect just failing outright. I think the only errors I've seen are after a successful connection.

codetheweb:
I'm starting to see the value of testing against a stub implementation (working on a package that mocks a Tuya device).

It appears that since the connect function didn't wait for the get function to finish before returning, it's possible for data going to the socket/device to be overwritten.  Just pushed out a patch for it in v4.0.2, may fix some issues.

codetheweb:
The initial version of the mock testing library has now [been published](https://www.npmjs.com/package/@tuyapi/stub).

I found a few more bugs just developing it, they've been fixed in v4.0.3 of TuyAPI.

PR #153, opened 2019-02-21
Cleaned up _encode implementation using one buffer to assemble payloa…

kueblc:
Use one buffer to assemble payload
Remove string representations of binary data


kueblc:
@codetheweb Any special reason for implementing the encoder and parser as Class objects with static wrappers? Could skip the middle man and put encode and parse logic directly in static function.

PR #160, opened 2019-02-26
Parser cipher updates

kueblc:
I have made conservative revisions to the parser and cipher components as per our discussions. Line count may be high due to indentation change. There should no be any breaking changes for user facing interfaces in this PR, aside from MessageParser class.

The highlights are:
- Reduced external dependencies by utilizing built in `crypto` module and (re)moving unused modules
- MessageParser class with static wrappers replaced with static functions
- Fixed parsing bugs related to return code
- More robust and consistent error detection in parsing
- Coverage of these components brought to 100%

I wanted to get your input at this point, as the application logic and division of responsibilities remain the same and changes should go mostly unnoticed by users. I did not touch index.js in this PR so a few changes I'm working on have been held back.

Moving forward I'd like to decouple responsibilities between decoding/encoding frames, decrypting/encrypting data, and decoding/encoding JSON. This should allow us to have a more consistent interface for developers, improve cohesion, while still leaving the user facing API untouched.

I'd also like to add CRC generation and checking, parser handling of multi message buffers, as well as adding other available data to the parsed message object.


codetheweb:
Thanks, this looks great.  Very cool you were able to get rid of the `node-forge` dependency as that package is huge and time consuming to install.

I can't believe I accidentally installed `clone` to the package dependencies. 🤦‍♂️  

What do you mean by

> Moving forward I'd like to decouple responsibilities between decoding/encoding frames, decrypting/encrypting data, and decoding/encoding JSON.

?

Do you mean have a separate interface/function for, say, encoding JSON and another for encoding a raw buffer?

> I'd also like to add CRC generation and checking, parser handling of multi message buffers, as well as adding other available data to the parsed message object.

CRC generation was added at one point, but I removed it because it didn't match the CRC algorithm used by the device.  If you want to add it that's great but it doesn't really seem to matter one way or another.

Merging.

codetheweb:
By the way, do you want me to add you as a contributor to this repository so you can push and pull to a branch directly in this repo?

kueblc:
> Thanks, this looks great. Very cool you were able to get rid of the `node-forge` dependency as that package is huge and time consuming to install.

You're welcome! I'm glad the revisions are to your liking.

> What do you mean by
> 
> > Moving forward I'd like to decouple responsibilities between decoding/encoding frames, decrypting/encrypting data, and decoding/encoding JSON.
> 
> ?
> 
> Do you mean have a separate interface/function for, say, encoding JSON and another for encoding a raw buffer?

I noticed several places where messages are cast between objects, JSON encoded strings, base64 encoded strings, buffers, etc. There are also some assumptions made in checking for encryption which also happens in several places. I think we could eliminate a lot of branching by strictly defining what types the parser and encoder accept and return, and handling when to encode/decode encrypt/decrypt at the application level.

While the current model might seem more developer friendly by offering a lot of options for what an interface accepts and returns, I'd argue that it's easier to read an application when the steps are laid out explicitly. For instance, rather than having an option to output encryption as a buffer or base64, let's just say (for example) we only output as base64, which is probably what I want. In the edge case that I do want a buffer instead, I only need to add one line `Buffer.from(encryptedString, 'base64');`.

Again I'm not the best at explaining what I mean, probably better understood if I can provide the code for you to compare :stuck_out_tongue: 

> > I'd also like to add CRC generation and checking, parser handling of multi message buffers, as well as adding other available data to the parsed message object.
> 
> CRC generation was added at one point, but I removed it because it didn't match the CRC algorithm used by the device. If you want to add it that's great but it doesn't really seem to matter one way or another.

I noticed that Tuya's implementation of CRC is a little weird (read: wrong) but I have a matching implementation working https://github.com/kueblc/mocktuyacloud/blob/master/lib/crc.js. Sure, the devices don't seem to care much now, but it's proven to be a reliable check in my home automation setup, and at a cost of ~3kb I think it's worth it.

Actually, funny enough, running the tuyapi tests fires off CRC mismatch warnings on my setup due to my device detection component trying to add the fake tuyastub :laughing: 

> By the way, do you want me to add you as a contributor to this repository so you can push and pull to a branch directly in this repo?

That would be swell! How do you want to communicate on revisions moving forward?

codetheweb:
You should have access now.  If not, try clicking [this link](https://github.com/codetheweb/tuyapi/invitations).

I was thinking we would make changes on `feature-*` or `dev-*` branches, then open PRs to merge them into `development`.

PR #162, opened 2019-02-26
Terse defaults

kueblc:
In this PR, I've replaced manual argument checking with ES6 style defaults. I personally feel this is more readable but I'll await your input before proceeding.


codetheweb:
Looks great, thanks!

PR #178, opened 2019-03-24
Implement Tuya flavored CRC

kueblc:
Requires `tuyapi/stub` to use the development version of `message-parser` in order to pass tests, due to CRC previously being ignored.


codetheweb:
Great work as always, thank you.

Do you think we should merge development to master at this point so tests continue to work?

kueblc:
> Do you think we should merge development to master at this point so tests continue to work?

Not just yet, I made a bad commit somewhere I have to track down. I think I can take care of it tomorrow.

PR #182, opened 2019-03-31
Fix CRC computation

kueblc:
I made a goof! Accidentally used the wrong byte-range for CRC computation. Fixed now, sorry about the hasty PR, will try to be better about thorough testing.

Merge with master whenever you're ready and tests *should* pass.


codetheweb:
👍

PR #183, opened 2019-04-01
Slight refactoring, add proper CRC support

codetheweb:


codetheweb:
Just merged @kueblc and the tests aren't passing on my machine after publishing a new version of TuyAPI (v4.0.4), bumping TuyAPI's version in @tuaypi/stub, and linking both to test locally.

I'll have to look at it more tomorrow.

Issue #176, opened 2019-03-21
Overwriting socket data / not waiting for a response


codetheweb:
Discovered this issue when working on a separate project that depends on TuyAPI.

(I believe this may be the root cause of https://github.com/codetheweb/tuyapi/issues/158 and https://github.com/codetheweb/tuyapi/issues/103.)

I added a condition to the data event handler for the socket that checks if the parsed and returned data is an object; if not it would throw an error.  After adding this, tests started failing.

I discovered that on a fast enough machine (and testing against a local stub) it's possible to overwrite data sent to the socket by calling `.set()` and/or `.get()` quickly after each other.  Data has already been being overwritten in the test suite, but it wasn't caught because I didn't add a check to see if the returned data was an object or not.

For instance, running the test suite generated this beautiful packet:

```
000055aa000000000000000a000000607b226465764964223a223232333235313836646234613232313764633865222c2267774964223a223232333235313836646234613232313764633865222c22647073223a7b2231223a747275652c2232223a747275657d7d000000000000aa55000055aa000000000000000a000000617b226465764964223a223232333235313836646234613232313764633865222c2267774964223a223232333235313836646234613232313764633865222c22647073223a7b2231223a747275652c2232223a66616c73657d7d000000000000aa55000055aa000000000000000a000000617b226465764964223a223232333235313836646234613232313764633865222c2267774964223a223232333235313836646234613232313764633865222c22647073223a7b2231223a747275652c2232223a66616c73657d7d000000000000aa55
```

~~Obviously invalid~~.  I took a stab at implementing a [function queue system](https://github.com/codetheweb/tuyapi/tree/patch-response-bad-data), but it's not working.

Thoughts @tjfontaine @kueblc @unparagoned?

kueblc:
I don't think that beautiful packet is necessarily invalid; I have seen devices send multiple TuyaFrames in a single packet, I've handled it [like so](https://github.com/kueblc/mocktuyacloud/blob/master/lib/lan-frame.js#L33).

kueblc:
I haven't had the chance to do a thorough investigation, forgive me if I'm misunderstanding the issue. Breaking the new firmware has been keeping me quite occupied. I'll circle back.

codetheweb:
No problem.

That's some very elegant code, I like it. :)

I guess it's not necessarily an invalid packet, but adding parsing abilities for that would be a bit complex if we still want to support synchronous-style code.  For instance, a user might have a snippet that looks something like this:

[...]

PR #184, opened 2019-04-01
Same bug as #182 but for encode instead of parse

kueblc:
I give up :stuck_out_tongue_closed_eyes: 
Please forgive me
Didn't catch this when testing with real devices because as you already know they don't enforce CRC
Tests now pass

### DIFF
--- a/lib/message-parser.js
+++ b/lib/message-parser.js
@@ -136,7 +136,7 @@ function encode(options) {
 
   // Add payload, crc, and suffix
   payload.copy(buffer, 16);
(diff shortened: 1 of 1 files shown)


codetheweb:
👍 everything appears to be fixed except for cases when the stub device emits a packet with two responses.  It feels wrong, but they'll *probably* pass on Travis due to running more slowly than they do on my local machine.  Still working on #176 to fix this (don't need a code review yet, but if you could comment there on what you've experienced with devices it'd be very helpful).

PR #188, opened 2019-04-06
Add support for multiple payloads in single packet, bump dependencies

codetheweb:
@kueblc mind looking this over?  Spent a while on this, it adds support for parsing multiple payloads in a single packet as well as using sequence numbers to use the correct callback for synchronous code.  The tests against @tuyapi/stub should pass, but I've been having a lot of issues using symlinks with `npm link` locally; so I haven't really been able to test.  It does work great against my physical smart plug, for what it's worth.

The changes to stub can be found [here](https://github.com/TuyaAPI/stub/compare/feature-sequence-n).

### DIFF
--- a/index.js
+++ b/index.js
@@ -66,6 +66,8 @@ class TuyaDevice extends EventEmitter {
 
     this._currentSequenceN = 0;
     this._resolvers = {};
+
+    this._waitingForSetToResolve = false;
   }
 
   /**
@@ -195,12 +197,13 @@ class TuyaDevice extends EventEmitter {
     });
 
     // Send request and wait for response
+    this._waitingForSetToResolve = true;
     return new Promise((resolve, reject) => {
       try {
         // Send request
-        this._send(buffer).then(data => {
-          resolve(data);
-        });
+        this._send(buffer);
+
+        this._setResolver = resolve;
       } catch (error) {
         reject(error);
       }
@@ -404,6 +407,17 @@ class TuyaDevice extends EventEmitter {
      * @property {Number} sequenceN the packet sequence number
      */
     this.emit('data', packet.payload, packet.commandByte, packet.sequenceN);
+
+    // Status response to SET command
+    if (packet.sequenceN === 0 &&
+        packet.commandByte === CommandType.STATUS &&
+        this._waitingForSetToResolve) {
+      this._setResolver(packet.payload);
+
+      // Remove resolver
+      this._setResolver = undefined;
+      return;
+    }
 
     // Call data resolver for sequence number
     if (packet.sequenceN in this._resolvers) {
--- a/index.js
+++ b/index.js
@@ -387,6 +387,9 @@ class TuyaDevice extends EventEmitter {
 
     if (packet.commandByte === CommandType.HEART_BEAT) {
       debug(`Pong from ${this.device.ip}`);
+
+      // Remove resolver
+      delete this._resolvers[packet.sequenceN];
       return;
     }
 
@@ -404,14 +407,6 @@ class TuyaDevice extends EventEmitter {
     // Call data resolver for sequence number
     if (this._resolvers[packet.sequenceN]) {
       this._resolvers[packet.sequenceN](packet.payload);
-
-      // Remove resolver
-      delete this._resolvers[packet.sequenceN];
-    } else if (packet.sequenceN === 0) {
-      // Returned sequence number is 0, probably a response to a set command
-
-      // Call the first resolver in the queue
-      this._resolvers[Object.keys(this._resolvers)[0]](packet.payload);
 
       // Remove resolver
       delete this._resolvers[packet.sequenceN];
@@ -505,8 +500,6 @@ class TuyaDevice extends EventEmitter {
 
         const thisID = dataRes.payload.gwId;
         const thisIP = dataRes.payload.ip;
-
-        console.log(this.foundDevices);
 
         // Add to array if it doesn't exist
         if (!this.foundDevices.some(e => (e.id === thisID && e.ip === thisIP))) {
--- a/index.js
+++ b/index.js
@@ -247,7 +247,8 @@ class TuyaDevice extends EventEmitter {
     // Create byte buffer
     const buffer = this.device.parser.encode({
       data: Buffer.allocUnsafe(0),
-      commandByte: CommandType.HEART_BEAT
+      commandByte: CommandType.HEART_BEAT,
+      sequenceN: ++this._currentSequenceN
     });
 
     // Send ping
@@ -405,7 +406,7 @@ class TuyaDevice extends EventEmitter {
     this.emit('data', packet.payload, packet.commandByte, packet.sequenceN);
 
     // Call data resolver for sequence number
-    if (this._resolvers[packet.sequenceN]) {
+    if (packet.sequenceN in this._resolvers) {
       this._resolvers[packet.sequenceN](packet.payload);
 
       // Remove resolver
--- a/package-lock.json
+++ b/package-lock.json
@@ -1203,7 +1203,9 @@
       "dev": true
     },
     "@tuyapi/stub": {
-      "version": "0.1.2",
+      "version": "0.1.3",
+      "resolved": "https://registry.npmjs.org/@tuyapi/stub/-/stub-0.1.3.tgz",
+      "integrity": "sha512-Jd1bmGgJNd94h3o7/fQjctikphKZxp6FyJ3UTpNFZ2nJcN9GcGlpC+Zi2vTE9OcW1A+FFfPMT82WagFNqH+7mg==",
       "dev": true,
       "requires": {
         "debug": "^4.1.1",
@@ -2627,6 +2629,12 @@
           "dev": true
         }
       }
+    },
+    "delay": {
+      "version": "4.2.0",
+      "resolved": "https://registry.npmjs.org/delay/-/delay-4.2.0.tgz",
+      "integrity": "sha512-EBX+pZE4qSowGAMr6M0cLiPRQu2Kus/qTNLO7c+EoXpTPJH9ApFdHX+cQU1WsSHXgwhLyidfZ5Hxuq6ctWhSdw==",
+      "dev": true
     },
     "delayed-stream": {
       "version": "1.0.0",
@@ -9718,6 +9726,28 @@
       "dev": true,
       "requires": {
         "safe-buffer": "^5.0.1"
+      }
+    },
+    "tuyapi": {
+      "version": "4.0.4",
+      "resolved": "https://registry.npmjs.org/tuyapi/-/tuyapi-4.0.4.tgz",
+      "integrity": "sha512-0gU6um3Imj3jHNm1cUuP1mXJdf0Z7H+kjhqlQfxuqUwwaB4hAmENatF6mmRhpF6NDzq181rXxU89hJpmw37Lmg==",
+      "dev": true,
+      "requires": {
+        "debug": "4.1.1",
+        "p-retry": "4.1.0",
+        "p-timeout": "3.0.0"
+      },
+      "dependencies": {
+        "p-timeout": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/p-timeout/-/p-timeout-3.0.0.tgz",
+          "integrity": "sha512-HKUsVzU+2A+CcItUxgZ4Q1th5Hh2DHtSsh7gLTMkrL8Ki4Ss736nFp+yqb9M/ZKSKb5il0IXeLzBmUqD3k3mzQ==",
+          "dev": true,
+          "requires": {
+            "p-finally": "^1.0.0"
+          }
+        }
       }
     },
     "tweetnacl": {
--- a/package.json
+++ b/package.json
@@ -42,10 +42,11 @@
     "p-timeout": "3.1.0"
   },
   "devDependencies": {
-    "@tuyapi/stub": "0.1.2",
+    "@tuyapi/stub": "0.1.3",
     "ava": "1.4.1",
     "clone": "2.1.2",
     "coveralls": "3.0.3",
+    "delay": "4.2.0",
     "documentation": "9.3.1",
     "nyc": "13.3.0",
     "xo": "0.24.0"
--- a/test/find.js
+++ b/test/find.js
@@ -1,12 +1,20 @@
 import test from 'ava';
 import TuyaStub from '@tuyapi/stub';
 import clone from 'clone';
+import delay from 'delay';
 
 const TuyAPI = require('..');
 
 const stub = new TuyaStub({id: '22325186db4a2217dc8e',
                            key: '4226aa407d5c1e2b',
                            state: {1: false, 2: true}});
+
+// You may notice that at the end of each test
+// there's a delay() before the function exits.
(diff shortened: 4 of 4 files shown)


kueblc:
Looking good! Haven't had a chance to test.

I like that by using sequence numbers to track responses, you can also get useful data in the promise callback.

Question, are you reverting to a class pattern in message-parser?

codetheweb:
Yes, because the MessageParser now implements decrypting message payloads if they're encrypted with the Cipher class (for which it needs the local key and thus it makes sense to have a MessageParser class).

I know you're not a fan of the all-in-on decoding function; maybe we could add additional functions that only did one part of the decoding process to make it more modular?

Haven't had time to add documentation yet.

codetheweb:
Yeah, I want to merge into master so we can publish a new version of TuyAPI, then publish a new version of @tuyapi/stub with dependencies bumped so tests work again.

codetheweb:
Finally managed to get `npm link` working; tests now pass if you use the `feature-sequence-n` branch of @tuyapi/stub.

Also added documentation.

codetheweb:
Alright, actually fixed now.

The issue was the CRC was being calculated before the sequence number was written to the buffer.

codetheweb:
Just a friendly ping @kueblc.  Any time you can take to look at this is much appreciated.

kueblc:
@codetheweb Thanks for the reminder - I've put it on my schedule for tomorrow

index.js
```diff
@@ -171,7 +150,7 @@ class TuyaDevice extends EventEmitter {
    *             '1': true,
    *             '2': 'white'
    *          }}).then(() => console.log('device was changed'))
-   * @returns {Promise<Boolean>} - returns `true` if the command succeeded
+   * @returns {Promise<Object>} - returns response from device
```
kueblc: Love this, far more useful return. This may be a breaking change if folks have written their scripts to check for a boolean response. Consider version bump?

index.js
```diff
@@ -210,41 +188,20 @@ class TuyaDevice extends EventEmitter {
     debug('SET Payload:');
     debug(payload);
 
-    // Encrypt data
-    const data = this.device.cipher.encrypt({
-      data: JSON.stringify(payload)
-    });
-
-    // Create MD5 signature
-    const md5 = this.device.cipher.md5('data=' + data +
```
kueblc: Glad you found a better home for this. Index is a little bloated as is.

index.js
```diff
@@ -210,41 +188,20 @@ class TuyaDevice extends EventEmitter {
...
-    const thisData = Buffer.from(this.device.version + md5 + data);
-
     // Encode into packet
-    const buffer = Parser.encode({
-      data: thisData,
-      commandByte: 7 // 0x07
+    const buffer = this.device.parser.encode({
+      data: payload,
+      encrypted: true, // Set commands must be encrypted
+      commandByte: 7, // 0x07
```
kueblc: Could we introduce an "enum" ala https://github.com/kueblc/mocktuyacloud/blob/master/lib/lan-frame.js#L2 for readability and convenience? Could also use `const`s.

index.js
```diff
@@ -335,53 +294,24 @@ class TuyaDevice extends EventEmitter {
...
-            }
-
-            if (dataRes.commandByte === 0x07) { // Set succeeded
-              debug('Set succeeded.');
-              return;
-            }
-
-            debug(`Undefined data with command byte ${dataRes.commandByte}`);
-          } else { // Message is encrypted
-            data = this.device.cipher.decrypt(data);
```
kueblc: Really happy with this direction, glad you could find a better home for crypto functions

index.js
```diff
@@ -453,6 +383,44 @@ class TuyaDevice extends EventEmitter {
...
+    this.emit('data', packet.payload, packet.commandByte, packet.sequenceN);
+
+    // Call data resolver for sequence number
+    if (this._resolvers[packet.sequenceN]) {
+      this._resolvers[packet.sequenceN](packet.payload);
+
+      // Remove resolver
+      delete this._resolvers[packet.sequenceN];
+    } else if (packet.sequenceN === 0) {
+      this._resolvers[Object.keys(this._resolvers)[0]](packet.payload);
```
kueblc: Not sure I get this, without looking into it too much.

index.js
```diff
@@ -453,6 +383,44 @@ class TuyaDevice extends EventEmitter {
...
+     * @property {Number} commandByte
+     * commandByte of result
+     * (e.g. 7=requested response, 8=proactive update from device)
+     * @property {Number} sequenceN the packet sequence number
+     */
+    this.emit('data', packet.payload, packet.commandByte, packet.sequenceN);
+
+    // Call data resolver for sequence number
+    if (this._resolvers[packet.sequenceN]) {
+      this._resolvers[packet.sequenceN](packet.payload);
```
kueblc: LGTM

index.js
```diff
@@ -492,11 +460,25 @@ class TuyaDevice extends EventEmitter {
...
    */
   checkIfValidString(input) {
-    if (input === undefined || typeof input !== typeof 'string' || input.length === 0) {
-      return false;
-    }
+    return typeof input === 'string' && input.length > 0;
+  }
 
-    return true;
+  arrayDeepInclude(arr, obj) {
```
kueblc: At this point we might consider adding a `utils.js`

index.js
```diff
@@ -492,11 +460,25 @@ class TuyaDevice extends EventEmitter {
...
+        return false;
+      }
+
+      return true;
+    });
+
+    return !result;
+  }
+
+  wrapFunction(fn, context, params) {
```
kueblc: `util`

index.js
```diff
@@ -492,11 +460,25 @@ class TuyaDevice extends EventEmitter {
    * `true` if is string and length != 0, `false` otherwise.
    */
   checkIfValidString(input) {
```
kueblc: `util`

lib/cipher.js
```diff
@@ -45,24 +45,23 @@ class TuyaCipher {
   * returns object if data is JSON, else returns string
   */
   decrypt(data) {
+    // Incoming data format
+    let format = 'buffer';
+
     if (data.indexOf(this.version.toString()) !== -1) {
```
kueblc: Although not part of your recent changes, I noticed this during the review. The version string should always be the first 3 characters, so I recommend `if (data.startsWith(this.version)) {`

lib/cipher.js
```diff
@@ -12,8 +11,7 @@ const forge = require('node-forge');
 */
 class TuyaCipher {
   constructor(options) {
-    this.cipher = forge.cipher.createCipher('AES-ECB', options.key);
-    this.decipher = forge.cipher.createDecipher('AES-ECB', options.key);
+    this.key = options.key;
     this.version = options.version;
```
kueblc: `options.version.toString()` so we don't have to convert it everytime

lib/message-parser.js
```diff
@@ -1,135 +1,222 @@
...
+    const expectedCrc = buffer.readInt32BE(HEADER_SIZE + payloadSize - 8);
+    const computedCrc = crc(buffer.slice(0, payloadSize + 8));
+
+    if (expectedCrc !== computedCrc) {
+      // eslint-disable-next-line max-len
+      throw new Error(`CRC mismatch: expected ${expectedCrc}, was ${computedCrc}. ${buffer.toString('hex')}`);
     }
 
-    return true;
+    return {payload, leftover, commandByte, sequenceN};
```
kueblc: Very nice :stuck_out_tongue:

lib/message-parser.js
```diff
@@ -139,55 +226,45 @@ class MessageParser {
...
-  return p._decode();
-}
+  /**
+   * Given a complete packet, writes
+   * a sequence number into the packet.
+   * @param {Buffer} buffer packet
+   * @param {Number} n sequence number
+   * @returns {Buffer}
+   */
+  writeSequenceN(buffer, n) {
```
kueblc: Why is this separate? Should be able to encode a message with the complete set of parameters. Use case where you might not?

Seems inconsistent with the other parameters of the message. If we want to construct a message piecemeal we should be able to set each parameter. But my preference would be enforcing a pattern where the message isn't encoded until you have all the parameters.

codetheweb: This was an artifact from when I was testing other stuff out, removed.

lib/cipher.js
codetheweb: Yes, but the function accepts Buffers as well as strings. Buffers don't have a `.startsWith()` function.  Updated the documentation to make that more clear.

index.js
codetheweb: Forgive me for not understanding, but why would a structure like that be better? You'd still have to set the same properties.

index.js
codetheweb: As long as users have done something like 
```javascript
if (result) {
  console.log('Worked!');
}
```

or 

```javascript
if (!result) {
  console.error('Error!');
}
```

and not

```javascript
if (result == true) {
  console.log('Worked!');
}
```

they should be fine.

It's technically a breaking change, so according to SemVer we should increment.  But I dunno, it's not like we've added any new features.

index.js
codetheweb: Added some more comments; hopefully it's clearer now.

codetheweb:
Thanks for all the great feedback @kueblc.

Incorporated most suggestions and commented on a couple others.

lib/cipher.js
kueblc: Ah yes, didn't notice it took Buffers as well. In that case, it should be `if (data.indexOf(this.version.toString()) === 0) {`.

index.js
kueblc: It wouldn't make a functional difference but IMO it would be easier on the programmer to understand the significance of the values `0x07`, `0x0a` etc, and also provide a reference as to the available commands.

The same line could read `commandByte: CommandType.CONTROL,` and then its immediately obvious what the property means. You also have a sense of the other values you could pass to activate other functions of the device, ie, performing device upgrades or relocating the device to a new WiFi network.

index.js
kueblc: That's the case I'm noting, yes technically semver would specify a version bump, and yes it's a "small" breaking change, so I don't think it would be terrible. On the other hand, version bump doesn't really "cost" anything. Ultimately your call, I'm fine either way.

index.js
kueblc: Forgive me, I'm still trying to follow the logic. What's the significance of the first resolver? What if there are no resolvers ie `Object.keys(this._resolvers)[0]` is `undefined`?

index.js
codetheweb: Ok, that makes more sense now.  Is it ok if I copy the list of control bytes from mocktuyacloud?

index.js
codetheweb: > What's the significance of the first resolver?

There is none; but when TuyAPI receives a response with the sequence number set to 0 it needs to resolve to *something*.

> What if there are no resolvers ie Object.keys(this._resolvers)[0] is undefined?

Good point, not sure what should happen.  That shouldn't occur during normal operation but it probably would in some edge case.

index.js
codetheweb: Ok, I'll bump the major version just to keep it consistent.

index.js
kueblc: Absolutely, go for it

index.js
kueblc: Do you mind if I take a day or two to wrap my head around this? I'll have a little time to experiment with this during the weekend.

codetheweb: Sure, no rush.

index.js
codetheweb: Added the enum, it does look a lot better now.

index.js
kueblc: Oops, sorry "a day or two" turned into so much more! Let's get back to it, shall we?

I took some time to think about this, and I don't think this logic is quite right.

If you issue a set command (0x07) you'll get an acknowledge response (0x07) with the same sequence number, then subsequently a status update (0x08) with sequence 0.

So this means with each set command, this logic eats two resolvers, the second one being whatever is next, not necessarily related.

What we really want is for the Promise to resolve once for each sequence number. In the event that we don't have a matching sequence number in our resolvers, the Promise has already been resolved via the acknowledge message, or is some other kind of message or garbage data we don't care about.

So I think the correct thing to do here is actually nothing. That is, remove the `else if` block, and I think we're golden.

kueblc: Also, consider that we haven't really lost any information if we don't resolve on status update (0x08) since the `data` event is still fired.

kueblc: I might also change `if (this._resolvers[packet.sequenceN])` to `if (packet.sequenceN in this._resolvers)` to make it explicit that we are checking if the member exists, not whether the member is truthy. Tiny detail though.

kueblc: I noticed that we're also adding a Promise for the heartbeat pings (at `this._resolvers[0]`), but never resolve them due to the logic in packet handler that filters them out.

Perhaps allow the heartbeat packet to resolve 0, and move the heartbeat response handling to `_sendPing` to match the logic of `get` and `set`.

codetheweb: > So I think the correct thing to do here is actually nothing. That is, remove the else if block, and I think we're golden.

Makes sense now that you explained it like that.

> Also, consider that we haven't really lost any information if we don't resolve on status update (0x08) since the data event is still fired.

True, but if the user is using a Promise/callback style, they don't get the result back.  So I think we should always resolve if the response sequence number has been mapped to a resolver.

> I noticed that we're also adding a Promise for the heartbeat pings (at `this._resolvers[0]`), but never resolve them due to the logic in packet handler that filters them out.

I think `await`ing `this._send()` on line 254 is sufficent for handling the response.

Good catch, though, I added something to delete the heartbeat resolver in the packet handler.

Let me know what you think.


Tests against stub are not currently passing; I'll take a closer look tonight when I have access to a physical device.

index.js
```diff
@@ -387,6 +387,9 @@ class TuyaDevice extends EventEmitter {
 
     if (packet.commandByte === CommandType.HEART_BEAT) {
       debug(`Pong from ${this.device.ip}`);
+
+      // Remove resolver
+      delete this._resolvers[packet.sequenceN];
```
kueblc: Ah yes, but `HEART_BEAT` packets have not been assigned a `sequenceN`, so we're still deleting `this._resolvers[0]` every time.

Are you sure there are no cases when two `HEART_BEAT` packets would arrive? Could network lag, device failure, or a malicious actor create this situation?

What about giving each packet a unique `sequenceN`? No special treatment for `HEART_BEAT` packets.

To clarify, what do you think about [something along these lines](https://github.com/kueblc/tuyapi/commit/a3299cb7cd9ca006d0b4da0aa1a97d41b8ab917a)? I left the `HEART_BEAT` debug message, along with the early `return` so no `data` event is emitted for `HEART_BEAT`, though this could be moved too. I suppose most use cases shouldn't need to know about these anyway.

codetheweb: I added your changes, makes sense.

The only issue left is that there's no data attached to the resolver for the `set()` calls now.

What do you think about having a separate resolver property for set commands?

codetheweb:
@kueblc I fixed it, it seems to work good but if you want to look it over that'd be great.

Just merged into `development`, going to merge into `master` and then release.

codetheweb:
It was a bit messy @kueblc but I merged everything locally and force-pushed and it seems to be good now.

Issue #175, opened 2019-03-14
Tuya 2.0 Firmware


tsightler:
**Describe the bug**
I have about 18 generic tuya wall switches/plugs that have always worked just fine with tuyapi, however, a couple of those switches recently received update to 2.0.0 firmware (all others are on 1.0.4 to 1.0.5).  After this I can no longer seem to use tuyapi to communicate with these devices, I just get socket error.

I can still use tuya-cli with list-app to get device ID/key, but any attempt to communicate with the devices using this new firmware with the discovered id/key simply fails with "Error from socket" message.

Any chance this is a known compatibility issue or is there any data I can provide that might help isolate the issue further?  I'm thinking I may need to try to block these things from Tuya so that they won't get future firmware updates.  Either that, or just suck it up and switch them to custom firmware.

kueblc:
Thanks for the report, I've been aware that an update was coming but the extent of the damage is unknown. This is very helpful for evaluating what has changed.

Was the update elective or did it occur silently?

Are you able to get a copy of the firmware bin? And/or a network capture of an update?

You can block Tuya from pushing further updates if you intercept DNS requests for `a.gw.tuyaus.com` `a.tuyaus.com` and `a3.tuyaus.com` (replace `us` with `eu` or `cn` if you're in EU or China). You can use `dnsmasq` on your router or home server and configuring your router to use this DNS first. Note this effectively removes any cloud capability.

tsightler:
I'll admit in this case the update was elective.  I was actually working with support to resolve an issue with these two devices due to the fact they were running firmware 1.0.0.  However, I was anticipating them pushing 1.0.5, like all of my other devices, not 2.0.0, although the update to resolve my issue.  That being said, I have received silent updates previously, from 1.0.2 to 1.0.5.

I've been hesitant to block connection to Tuya due to the fact that I do make use of the cloud integration with Google Assistant/Alexa, so I'd have to plan an alternative.  Not that I can't do that but everything works just fine now so it's becoming motivated enough to want to change all of that automation.

How do you get the firmware bin?  I'm assuming there is a tool.  Point me there and I'll be happy to grab it.  I'll probe at this think over the weekend and see if I can figure out what it's doing.

kueblc:
Out of curiosity, what was the original issue that 2.0.0 resolved for you?

[...]

PR #214, opened 2019-06-14
Protocol 3.3 and encrypted discovery

kueblc:
Preliminary support for protocol 3.3, including support for the new UDP broadcast mechanism.

Written to prioritize minimal change (hacky) and no breaking API changes. To be refactored in v6.0.0.

Tests pass. Additional testing was done with simulated 3.1 and 3.3 devices.

It wouldn't hurt to get a few more in vivo tests before merge. @Apollon77 @tsightler @kezakez @EricPalmquist @sylwester- @ChristianHoefer @thirug010

### DIFF
--- a/index.js
+++ b/index.js
@@ -411,7 +411,8 @@ class TuyaDevice extends EventEmitter {
     // Status response to SET command
     if (packet.sequenceN === 0 &&
         packet.commandByte === CommandType.STATUS &&
-        this._waitingForSetToResolve) {
+        this._waitingForSetToResolve &&
+        typeof this._setResolver === 'function') {
       this._setResolver(packet.payload);
 
       // Remove resolver
--- a/index.js
+++ b/index.js
@@ -105,7 +105,6 @@ class TuyaDevice extends EventEmitter {
       commandByte: CommandType.DP_QUERY,
       sequenceN: ++this._currentSequenceN
     });
-    debug('sending', buffer.toString('hex'));
 
     // Send request and parse response
     return new Promise((resolve, reject) => {
--- a/index.js
+++ b/index.js
@@ -594,18 +594,14 @@ class TuyaDevice extends EventEmitter {
   async toggle(property = '1') {
     property = property.toString();
 
-    try {
-      // Get status
-      const status = await this.get({dps: property});
+    // Get status
+    const status = await this.get({dps: property});
 
-      // Set to opposite
-      await this.set({set: !status, dps: property});
+    // Set to opposite
+    await this.set({set: !status, dps: property});
 
-      // Return new status
-      return await this.get({dps: property});
-    } catch (error) {
-      throw error;
-    }
+    // Return new status
+    return this.get({dps: property});
   }
 }
 
--- a/index.js
+++ b/index.js
@@ -105,6 +105,7 @@ class TuyaDevice extends EventEmitter {
       commandByte: CommandType.DP_QUERY,
       sequenceN: ++this._currentSequenceN
     });
+    debug('sending', buffer.toString('hex'));
 
     // Send request and parse response
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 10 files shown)


Apollon77:
Coool. I‘m currently on vacation in tunesia but can give it a try in 1-2 weeks. So maybe all the others are faster then me ... ;-(

Apollon77:
Code Review wise looks ok to me, but doc is missing for the new methods ;-)

I currently try to get such a device with new protocol so that I can test by myself too.

Apollon77:
One more: was not the port for the tcp connection for the new device firmware also different? Or am I wrong with this?

kueblc:
> Code Review wise looks ok to me, but doc is missing for the new methods ;-)

New methods should be considered internal for the time being, they will be refactored out in v6.0.0. Perhaps I can clarify this with an underscore?

> One more: was not the port for the tcp connection for the new device firmware also different? Or am I wrong with this?

It is the same, at least with the samples I have received. The UDP discovery process happens on a new port however.

Apollon77:
Underscore is a Good convention for „private“ ... but in fact to start only with them maybe is confusing. Other idea is an @private as doc for the method. But in fact the same. 
So all ok ;-) (my point of view)

Thanks for the tcp port info.

kueblc:
Actually turned out to be an easier refactor than I was anticipating, so I've taken out the protocol specific functions. Now the change set is isolated to payload treatment.

EricPalmquist:
I tested with my device today - an outdoor 3 socket power strip and can read and write to all three.  You might consider in your constructor to show that the version must be specified as a string - it shows as an integer, which does not work.  Excellent work though - I appreciate your hard work.

Apollon77:
Oh that it right. So we should adjust the check (use == ?) or add type conversion.

kueblc:
> You might consider in your constructor to show that the version must be specified as a string - it shows as an integer, which does not work.

Good catch, though if you're using `.find()` you don't need to specify the version.

Should be fixed now. `MessageParser` now ensures `version` is a string on construction.

> Oh that it right. So we should adjust the check (use == ?) or add type conversion.

Funny I had used the `==` operator, but the linter complained, so I took it out to get tests passing without realizing the consequence. This linter drives me crazy sometimes, it doesn't like my coding style and fights with me on legitimate language features :stuck_out_tongue:

thirug010:
Hi,
I have tested with 2 devices ( firmware 3.1.1 , 1.0.7) for both it is not finding the device with device id, with IP-Address it is able find it but it fails on 'data' with below error and it is not decrypted, Please let me know if i am missing something 

image

kueblc:
> Hi,
> I have tested with 2 devices ( firmware 3.1.1 , 1.0.7) for both it is not finding the device with device id, with IP-Address it is able find it but it fails on 'data' with below error and it is not decrypted, Please let me know if i am missing something

Hi @thirug010, could you please post the test code you are using? If you are able to capture your network traffic please post this as well.

lib/cipher.js
```diff
@@ -1,5 +1,7 @@
 const crypto = require('crypto');
 
+const UDP_KEY = crypto.createHash('md5').update('yGAdlopoPVldABfn', 'utf8').digest();
```
codetheweb: Let's abstract the key out of the logic; maybe put it in a file called `config.json` or something?

index.js
```diff
@@ -572,18 +599,14 @@ class TuyaDevice extends EventEmitter {
   async toggle(property = '1') {
     property = property.toString();
 
-    try {
-      // Get status
-      const status = await this.get({dps: property});
+    // Get status
+    const status = await this.get({dps: property});
```
codetheweb: Good catch, don't know why I put in a try/catch block there. 😛

index.js
```diff
@@ -492,66 +497,88 @@ class TuyaDevice extends EventEmitter {
...
+        // Change protocol version if necessary
+        if (this.device.version !== dataRes.payload.version) {
           this.device.version = dataRes.payload.version;
 
-          // Cleanup
-          listener.close();
-          listener.removeAllListeners();
-          resolve(true);
+          // Update the parser
+          this.device.parser = new MessageParser({
```
codetheweb: So if a user passes in the IP manually to the constructor and don't use `.find()`, they'll have to pass in `version: 3.3` as well, correct?

If so, we'll need to add something to the docs about that.

codetheweb:
Looks pretty good @kueblc; ready to merge once those changes are made.

I can't test it myself as I don't have any v3.3 devices, but I trust that everything works.

And if you give me some examples of bad linter rules, I can change the config. 😉

lib/cipher.js
kueblc: Sure no problem

index.js
kueblc: :+1:

index.js
kueblc: Yep, correct.

kueblc: Added a note to the README notes section

kueblc:
> Looks pretty good @kueblc; ready to merge once those changes are made.

Ready when you are :+1: 

> And if you give me some examples of bad linter rules, I can change the config. wink

Cool, let's revisit for v6.0.0. Got a bunch of notes to go over.

Apollon77:
I also currently try to get sich devices. Seems I need to purchase some :-(

It is not really about „bad“ Linter rules. In fact they are right and eg === is best practice also nimmt eyes but sometimes you want to have the „feature“ that == brings ;-) but in those rare cases you need to tell the Linter that you know what you did and to ignore it.

thirug010:
@kueblc,

Please find the network logs ([logs.zip](https://github.com/codetheweb/tuyapi/files/3294685/logs.zip)) and I use the basic code [tuya-dev.txt](https://github.com/codetheweb/tuyapi/files/3294686/tuya-dev.txt), Please let me know if you need any other information

Thanks 
Thiru

kueblc:
Thanks @thirug010, can you capture the network activity of the device? It looks like these are of your phone. You can use tcpdump or wireshark to create a pcap file.

Since `find` isn't working for you, I'd be interested to see what, if any, data is being broadcasted on UDP ports 6666-6668.

It also looks like you're passing `30` to `find`, which isn't valid

tsightler:
I tested the same code as @thirug010 (tuya-dev.txt) with my switches and the 5.1.0 tuya-api and they worked fine with both find and IP.

Unfortunately my primary use case for tuyapi is to use it with [tuya-mqtt](https://github.com/TheAgentK/tuya-mqtt) which currently doesn't work with tuya-api 5.x although it looks like it should be easy to modify for that part.  However, looks like I'll also need to modify it for supporting 3.3 firmware since that code currently requires specifying the IP for each device (no support for find) but has no method to specify firmware version.

AS137430:
@tsightler : https://github.com/TheAgentK/tuya-mqtt/blob/master/tuya-mqtt.js#L181 and https://github.com/codetheweb/tuyapi/issues/175#issuecomment-502475953.
Probably https://github.com/TheAgentK/tuya-mqtt/blob/master/tuya-mqtt.js#L125 needs to change too.

I have an (updated firmware) device new out from the box to test as well and have been waiting for @kueblc's changes. Thanks @kueblc and @codetheweb for your great work.

tsightler:
Yeah, I actually hacked up tuya-mqtt last night to not require the IP and just use the find() capability and it worked just fine with the newer firmware.  Thanks to everyone for all the work here, I'm happy to be able to return to local control for my devices.

AS137430:
@tsightler : Can you share your code please?

kueblc:
Thanks @tsightler that's great to hear it worked out well for you. Your data was instrumental in understanding the new protocol.

tsightler:
@AS137430 I will share the patched tuya-mqtt code in [my own tree](https://github.com/tsightler/tuya-mqtt) after flushing out a few remaining issues.  Mainly, while the code works to toggle the lights, it doesn't actually update the MQTT status as it seems like it's not correctly parsing the returned data.  I haven't had time to look at why but I suspect it's something pretty simple.

thirug010:
> Thanks @thirug010, can you capture the network activity of the device? It looks like these are of your phone. You can use tcpdump or wireshark to create a pcap file.
> 
> Since `find` isn't working for you, I'd be interested to see what, if any, data is being broadcasted on UDP ports 6666-6668.
> 
> It also looks like you're passing `30` to `find`, which isn't valid

@kueblc,
I have checked with new devices + different Tuya based app now it is working without any issues. 
I can confirm it works for firmware version 1.0.7, 3.1.1 and 3.3 without using the firmware versions in constructors ( works for both Tuya switch, dimmer and fan light dimmer) 

Thanks 
Thiru

tsightler:
@AS137430 I've published an initial version of tuya-mqtt that support 5.1.1 and discovery for 3.3 devices in my fork:  https://github.com/tsightler/tuya-mqtt
It seems to work for me but I only have simple, single switches from three different vendors and no lightbulbs so it's only tested in a small subset of possible use cases.  That being said, the overall changes are quite minimal.

PR #219, opened 2019-06-23
Fix crashing when the 'data' is a string

haimkastner:
From unknown reasons  (for me)  my curtain switches returning  'json obj data unvalid' in the first use, but after calling set with any value, they working properly and returning a valid status.

But the app crashes before I can set any value.

So I made the current PR change in my computer and it works for me.

About the 'json obj data unvalid' message, it happened to me only with the curtain switches (I have another 3 kinds of tuya switches) and one of the curtain switches running the 3.3 firmware version.

### DIFF
--- a/index.js
+++ b/index.js
@@ -111,7 +111,7 @@ class TuyaDevice extends EventEmitter {
       try {
         // Send request
         this._send(buffer).then(data => {
-          if (options.schema === true) {
+          if (typeof data !== 'object' || options.schema === true) {
             // Return whole response
             resolve(data);
           } else if (options.dps) {


kueblc:
This is actually two issues, that (1) `json obj unvalid` is returned from the device (indicating a problem with the `get` message) and the fact that (2) a device error breaks the functionality of `get` and potentially other functions.

For (2), this is great for a quick fix, but I think that ultimately, instead of returning the error message, we should fire an `error` event with the error message as part of the payload.

Any chance you could record the network activity when controlling the device by the app? This way we can hunt down the root cause of the error (1).

codetheweb:
I think I had this logic in place originally @haimkastner, maybe it got removed at some point.  Thanks for the catch.

@kueblc are you thinking we just check the return value for equality with `json obj data unvalid`?

kueblc:
> @kueblc are you thinking we just check the return value for equality with `json obj data unvalid`?

The return code should be non-zero for any error from the device, along with a plaintext payload describing (roughly) the error. So we should check for a non-zero return code in `MessageParser` or have it include the return code and we'll fire an error event from `TuyaDevice`.

haimkastner:
> Any chance you could record the network activity when controlling the device by the app? This way we can hunt down the root cause of the error (1).

Currently, the app communication with my devices works without a problem. I will try to factory-reset (long press on the main button, right?) my curtain device to see if the `json obj data unvalid` bug is returning.

And if so I will send you the Wireshark records.

kueblc:
> Currently, the app communication with my devices works without a problem. I will try to factory-reset (long press on the main button, right?) my curtain device to see if the `json obj data unvalid` bug is returning.

No we need a recording of the communication working correctly, so that we can fix the bug.

haimkastner:
OK, I recorded the current communication.
And when I changed the computer that talking with the devices the bug return.

This is the records (and the code I used)

[curain_records.zip](https://github.com/codetheweb/tuyapi/files/3321802/curain_records.zip)

kueblc:
This is the strangest thing, the `DP_QUERY` request is byte for byte identical, yet one returns the error and the other doesn't. Any ideas what might be different? I'll keep looking into it to see if I missed anything.

codetheweb:
Any update or is this not relevant anymore @haimkastner @kueblc?

kueblc:
Regarding https://github.com/codetheweb/tuyapi/pull/219#issuecomment-504756207

1) Ultimately I could not identify any difference between the recordings provided.

2) I might say we just merge this as another quick fix for now. Once one of us gets a chance we should fire an error when we get a non-zero return code, and reject the `Promise`.

haimkastner:
> Any update or is this not relevant anymore @haimkastner @kueblc?

Currently all works ok for me (with my work around).

Do you want that I retest it with the current version (5.1.2)?

codetheweb:
Sounds good.  I'll merge this in for now.

Added a note about this on the TuyAPI v6 [project](https://github.com/codetheweb/tuyapi/projects/2).

codetheweb:
Published in v5.1.3.

PR #232, opened 2019-07-27
use npx to run local installation

choongmanee:
generally, it's better practice to refrain from installing packages globally unless absolutely necessary to avoid cluttering commands in cli. using a local installation of `@tuyapi/cli` and `anyproxy` and executing them via npx will contain these packages to a project while providing the same results as executing globally installed packages in the project.

wasn't too sure to call this either a feature or bugfix. open to feedback. thanks

### DIFF
--- a/docs/SETUP.md
+++ b/docs/SETUP.md
@@ -1,5 +1,7 @@
 ## Linking a Tuya Device
 
+**requires npm@5.2 or greater**
+
 1. Add any devices you want to use with `tuyapi` to the Tuya Smart app.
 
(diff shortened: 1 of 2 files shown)


codetheweb:
In the future, please open an issue first so we can talk about any possible changes before you go through the trouble of making them.

In this case, installing `@tuyapi/cli` locally doesn't really make sense as the whole point of the package is to provide a global command line tool.  It does much more than just perform MITM attacks against the Tuya app.

I will be leaving the setup instructions as-is, let me know if you have any further questions or suggestions.

PR #292, opened 2020-02-16
Tuya Curtain switch

Hendriks046:
Hello,

Do you have any idea if it's possible to get a tuya curtain switch working in homebridge?

thanks already.

greets


codetheweb:
Please open an issue instead of a PR in the future.

There are several homebridge plugins available for Tuya, take a look: https://www.npmjs.com/search?q=homebridge%20tuya

PR #329, opened 2020-06-04
Fix callback that gets  never called on err

dotmaster:
Promise must be rejected in on('error') otherwise you cannot catch the error in the code.

### DIFF
--- a/index.js
+++ b/index.js
@@ -237,23 +237,23 @@ get(options = {}) {
     // Retry up to 5 times
   //  return pRetry(() => {
       return new Promise((resolve, reject) => {
-        const errorHandler =  (err) => {
+        const sendErrorHandler =  (err) => {
           debug('Error event from socket during _send.', this.device.ip, err);
 
           this.emit('error', new Error(`Error during _send from socket ${err} IP ${this.device.ip} device ${JSON.stringify(this.device)}`));
           this.client.destroy();
           reject(new Error(`Error during _send from socket ${err} IP ${this.device.ip} device ${JSON.stringify(this.device)}`));
-          this.client.removeListener('error', errorHandler);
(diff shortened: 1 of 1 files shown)


codetheweb:
Thanks for the contribution!

I sort of understand what you're going for, but I don't think it's going to work correctly as-is.  You're calling `reject()` for every single socket error, regardless of whether or not the error is related to the initial `connect()` call. A `catch()` statement like this would execute on errors not related to the connect:
```js
connect().catch(error => console.error(error))
```

dotmaster:
Hi, you are welcome! I am not quite sure if I understand your code completely, but it solved my issue I had with not being able to catch errors, when the device could not be reached through socket errors. E.g. I coult not catch Errors like these like these: 
Error from socket Error: connect ENETUNREACH 192.168.1.50:6668 - Local (0.0.0.0:59822)

In such a situation the code never resolved the promise when calling 
try {
  await device.connect();
}
catch(err){
  console.log("error in connect", err) //catch block gets never called
)

Instead I got an UnhandledPromiseRejectionWarning.

dotmaster:
Hi, I have looked into your code and start to understand the problem.
Makin Eventemitters use promises is kind of tricky and you need to keep track of the resolve functions.
Therefore on also need to register on('error') on various locations. 
I started now witch the functions connect and get, which are the ones I use.
The code ist starting to work now as I expect it to and I can catch errors from get and connect (tested both).
I haven´t refactured set yet as I wanted to have your opinion and probably you can d yourself.
What do you think?

codetheweb:
Hey @dotmaster are you still working on this or should I close it?

dotmaster:
It worked for me. So I would close it.

Max Isom <notifications@github.com> schrieb am So. 23. Aug. 2020 um 19:06:

PR #361, opened 2020-10-03
Handle null responses to get()

tsightler:
I know I'm supposed to submit to development, but that branch looks abandoned and not up to date with master.  I tried to follow the style guidelines and at least that test passes.

This is my attempt to improve the situation as it related to null responses to get requests.  I couldn't do anything with schema yet, as I've found no reliable alternative to get available DPS keys on devices that don't respond to DP_QUERY (still researching), however, I did implement fallback to using set=null for requests to get specific DPS keys.

I suppress the useless "json object unvalid" messages and instead try to return something that at least explains what happens (schema not supported or get failed to respond).  I suppress those messages being emitted to data as well.  Also got rid of the not usable null JSON messages.  I did consider changing get() to always return JSON and return an "error" object, but that might break existing stuff too much.

Added a timeout to set(), which was probably needed anyway, but I really needed now since a set=null to a non-valid DPS doesn't get a response to it hangs forever.

Got rid of what appeared to be useless _waitingForSetToResolve since it was set to true on the very first set call and never set to anything else. Looks like the code now depends on _setResolver being a function and maybe that was some holdover from a prior implementation.

This code now works great for me.  No useless "json object unvalid" messages returned to my app, no workarounds required on the user side to get DPS values, and I can just trap and handle the timeout messages.  I guess you could argue I should emit errors on timeouts as well, but honestly, what can I do about them really?  If I issue a set or a get, I'm almost certainly going to wait and act on that response so I think returning the messages as a data response is natural vs trying to deal with them through in error event.

Admittedly, I'm still only a novice at this Javascript stuff, so perhaps I'm talking crazy here, but I was just trying to get the code working reliable with all the devices I have and the devices my users have, so far, this seems to work and saves me from having to completely recode using the homebridge-tuya style implementation.  Thoughts and feedback are welcome.

I do plan to continue research on the newer devices and the commands the app uses, but it's unclear how successful I will be with that.

### DIFF
--- a/index.js
+++ b/index.js
@@ -74,7 +74,7 @@ class TuyaDevice extends EventEmitter {
     this._pingPongPeriod = 10; // Seconds
 
     this._currentSequenceN = 0;
-    this._resolvers = {};  
+    this._resolvers = {};
   }
 
   /**
@@ -96,7 +96,7 @@ class TuyaDevice extends EventEmitter {
    * tuya.get({schema: true}).then(data => console.log(data))
    * @returns {Promise<Boolean|Object>}
    * returns boolean if single property is requested, otherwise returns object of results
-   * or error message, 
+   * or error message
    */
   get(options = {}) {
     const payload = {
@@ -123,22 +123,23 @@ class TuyaDevice extends EventEmitter {
         // Send request
         this._send(buffer).then(async data => {
           if (data === 'json obj data unvalid') {
-             // Some devices don't respond to DP_QUERY so, for DPS get commands, fall back
-             // to using SEND with null value (so far, this appears to always work)
-             // For schema there's currently no fallback so provide a more clear data response
-            if (options.schema == true ) {
-              data = 'Schema for device not available'
+            // Some devices don't respond to DP_QUERY so, for DPS get commands, fall
+            // back to using SEND with null value (so far, this appears to always work)
+            // For schema there's currently no fallback so provide a clear response
+            if (options.schema === true) {
+              data = 'Schema for device not available';
             } else {
               const setOptions = {
                 dps: options.dps ? options.dps : 1,
                 set: null
-              }
-              data = await this.set(setOptions)
-              if (data = "No response to set request" ) {
-                data = "No response to get request"
+              };
+              data = await this.set(setOptions);
+              if (data === 'No response to set request') {
+                data = 'No response to get request';
               }
             }
           }
+
           if (typeof data !== 'object' || options.schema === true) {
             // Return whole response
             resolve(data);
@@ -185,7 +186,8 @@ class TuyaDevice extends EventEmitter {
    *           set: false,
    *           devId: '04314116cc50e346566e'
    *          }).then(() => console.log('device was turned off'))
-   * @returns {Promise<Object>} - returns response from device or timeout message on no response
+   * @returns {Promise<Object>} - returns response from device or timeout message
+   *  on no response
    */
   set(options) {
     // Check arguments
@@ -238,12 +240,12 @@ class TuyaDevice extends EventEmitter {
         this._send(buffer);
         this._setResolver = resolve;
         // Wait _responseTimeout seconds for _packetHandler to resolve set request
-        setTimeout(() => { 
+        setTimeout(() => {
           // If not resolved by packet handler resolve with timeout message
           if (typeof this._setResolver === 'function') {
             this._setResolver('No response to set request');
             this._setResolver = undefined;
-          }  
+          }
         }, this._responseTimeout * 2000);
       } catch (error) {
         reject(error);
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -128,6 +122,24 @@ class TuyaDevice extends EventEmitter {
       try {
         // Send request
         this._send(buffer).then(async data => {
+          if (data === 'json obj data unvalid') {
+            // Some devices don't respond to DP_QUERY so, for DPS get commands, fall
+            // back to using SEND with null value (so far, this appears to always work)
+            // For schema there's currently no fallback so provide a clear response
+            if (options.schema === true) {
+              data = 'Schema for device not available';
```
codetheweb, line 130: Instead of returning values in cases of an error, we should be throwing errors instead.

index.js
```diff
@@ -448,20 +448,22 @@ class TuyaDevice extends EventEmitter {
...
      * @event TuyaDevice#data
      * @property {Object} data received data
      * @property {Number} commandByte
      * commandByte of result
      * (e.g. 7=requested response, 8=proactive update from device)
      * @property {Number} sequenceN the packet sequence number
      */
-    this.emit('data', packet.payload, packet.commandByte, packet.sequenceN);
+    if (!(packet.commandByte === CommandType.DP_QUERY &&
+          packet.payload === 'json obj data unvalid')) {
```
codetheweb, line 460: There's a bunch of other possible error strings, so do we check for all of them here?

I feel like `data` should always be emitted regardless of what the received data actually means. At this point it's a design limitation; but since this is gonna be a breaking change anyways what do you think about adding another event (`state` or similar) that *only* emits when the device returns a valid JSON object (in addition to the current `data` event)?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tuyapi",
-  "version": "6.0.0",
+  "version": "6.0.1",
```
codetheweb, line 3: I would say this is technically a breaking change since `nullPayloadOnJSONError` was removed.

codetheweb (changes requested):

codetheweb:
Looks pretty good, thanks for the contributions.

index.js
tsightler: OK, but as a consumer, how do I associate the null response I receive with an error?  I mean, the get() call failed, it's not a null response from the device, it's a failure to get a response.  As a consumer of get() how can I know that the error emitted was the get call failed for this device?  I mean, I guess we could define an error strucure, with deviceId and request type so that the consumer would have some way to know that a given error was caused by a specific request.

However, my breaking API thought was to just make the response to get itself a JSON structure with a "response" property that contains the response and an optional "error" property if there is an error.  If we don't want to break existing we could add some type of flag to the get call to enable JSON responses.

index.js
tsightler: It's a valid point, but the reason I'm suppressing this one is because it's being caused by the API itself yet it's being surfaced to the API consumer.  As an API consumer I call get(), which ends up causing a data event sent to my consuming application.  So as that consumer what can I do with it?  It's not useful to me and my code didn't cause it (I just issued a get) so my upper level code has to suppress/ignore it.  Personally, I think it only makes sense to surface data from the device that is unsolicited (i.e. data update events) because everything else is useless to me as an API consumer anyway, so it's just noise.  Based on this I'd argue suppress all DP_QUERY responses since, as far as I know, those never come unsolicited from the device so they have to be a response to the API, not to the API consumer.

package.json
tsightler: Fair.  Personally, I want to break it a lot more!

index.js
codetheweb: > As an API consumer I call get(), which ends up causing a data event sent to my consuming application. So as that consumer what can I do with it? It's not useful to me and my code didn't cause it (I just issued a get) so my upper level code has to suppress/ignore it.

Absolutely, but some applications may want errors to be bubbled up. And for some errors (like when the timestamp field is wrong), it starts to get a little muddy on whether it's the package's fault or the consumer's fault.

> Based on this I'd argue suppress all DP_QUERY responses since, as far as I know, those never come unsolicited from the device so they have to be a response to the API, not to the API consumer.

I think you mean suppress everything **except** DP_QUERY responses, unless I misinterpreted the rest?

In any case, we can take this somewhere else. Probably outside the scope of this PR.

index.js
codetheweb: It might be more clear if we define what should happen in two cases:

- Consumer uses synchronous-style calls

Any error that is `throw`n from within a `get()` or `set()` function can be caught and handled.

- Consumer uses event-based calls

Any error that crops up can be emitted as an `error` event; however extra information is needed to associate it with a request. Exposing the sequence number within `error` events as well as returning it from calls to `get()/set()` would probably be the easiest way to handle that.

This is all making me depressed about the current API design. Maybe that'll be the motivation I need to actually knock out https://github.com/TuyaAPI/driver...

codetheweb: For now since it's possible for `data` events to contain strings anyways it should be fine.

tsightler:
OK, I'm going to retract this and resubmit with just the fallback changes.  I'll open another issue to discuss breaking changes for future versions.

PR #363, opened 2020-10-06
Fallback to set null when get response returns null

tsightler:
Here's an updated PR with only the minimal changes to implement a fallback to send() with null value when DP_QUERY returns the infamous "json obj data unvalid" which causes get() to return null value.

I had a user test a large variety of Tuya devices with this change and get() worked in all cases with this fix.  Unfortunately I don't have a fix for get schema yet, so that will still produce null on these devices with no special message, but we can at least document that behavior I guess.

Had to implement a simple timeout function on set() because a set() with null to a DPS that doesn't exist would hang forever since there was no response.

This should be non-breaking to any existing code, no text error messages, no suppression of the 'json obj data unvalid' response.

Thoughts welcome.

### DIFF
--- a/index.js
+++ b/index.js
@@ -247,6 +247,9 @@ class TuyaDevice extends EventEmitter {
         }
       });
     }).then(data => {
+      if (typeof this._setResolver === 'function') {
+        this._setResolver === undefined;
+      }
       return data;
     });
   }
--- a/index.js
+++ b/index.js
@@ -82,8 +82,7 @@ class TuyaDevice extends EventEmitter {
     this._currentSequenceN = 0;
     this._resolvers = {};
     this._setQueue = new PQueue({
-      concurrency: 1,
-      timeout: this._responseTimeout * 1000
+      concurrency: 1
     });
   }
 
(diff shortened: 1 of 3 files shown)


tsightler:
OK, this is too buggy.  This can still hang if you have mulitple outstanding set requests because _setResolver() is a single instance so it's either waiting for a resolve or not, but what happens if there are two outstanding set() requests?  The first one resolves, and the second one is never processed (easily repoducible by sending mulitple set request to non-exisitng DPS values on a device).  As a hackish workaround I've just made the timeout call resolve() in all cases, which I guess is better than hanging forever, and seems to work for most real world cases, but I'm not sure it's really acceptable.

codetheweb:
> This can still hang if you have mulitple outstanding set requests because _setResolver() is a single instance so it's either waiting for a resolve or not, but what happens if there are two outstanding set() requests?

Unfortunate side effect of having to support both synchronous and event based code. One workaround might be to add a 'lock' to the set function so that only one call can run at a time. I don't see any side effects of adding a lock, do you?

tsightler:
I was pretty much thinking of exactly this, but I wasn't sure if we'd need a queue to prevent races.  For example, what happens if I issue a set that is in the process of timing out, so the function is locked, however, during that time 3 other set request come in so are pending the lock.  What happens when the lock is released, is there any chance multiple concurrent requests sneak in before the function is relocked?  I was thinking we might need a simple queue to queue the set request whenever the function is locked.

Maybe I'm overthinking it though.

tsightler:
OK, take a look at this latest update which uses a different approach that I believe is the cleanest possible way with the least overall code changes.

Instead of using a lock I pulled in p-queue (hey, it's already using p-timeout and p-retry, so what's one more) and create a queue with concurrency 1 and a timeout.  Then I just wrapped the unchanged existing set() function, completely unchanged from the existing 6.0.0 code, inside the queue.add function.  This addresses both the lack of a timeout and the concurrency issue with just a handful of lines of code.

I've tested it pretty extensively as I had some reproducible hanging test cases, and it works for all of them every time.

codetheweb:
👍 looking good to me, only thing that needs to be fixed is `package.json`. Looks like some junk data sneaked in there and it was reformatted slightly.

tsightler:
Yeah, well, with more testing I realized I broke async calls because it now doesn't return a promise to the caller.  Stupid mistake on my part (I have a mix of sync and async code in my project, sync during initialization when I need to wait responses to make further decisions, async during normal function).  I need a promise wrapper around the queue.add, which I've already done locally, and it seems to work just fine, but starts making the set function ugly and hard to follow so I wanted get thoughts on the following approach:

1) Rename existing 6.0.0 set() to _set()
2) Create new set() which implements the queue calling _set()

To me this is cleaner and easier to follow as otherwise I end up with a bunch of nested promises which I think are difficult to follow.  I mean, technically the nested promises existing in both cases of course, but one why seems super clear to my novice eyes and the other way just gets me confused as heck.  Also open to other thoughts/suggestions.

codetheweb:
> Yeah, well, with more testing I realized I broke async calls because it now doesn't return a promise to the caller. Stupid mistake on my part (I have a mix of sync and async code in my project, sync during initialization when I need to wait responses to make further decisions, async during normal function).

Hmm, the automated tests should have caught this.

>  I need a promise wrapper around the queue.add, which I've already done locally, and it seems to work just fine, but starts making the set function ugly and hard to follow

Unless I'm missing something, it seems like you could just return `queue.add()` according to the documentation. Should resolve when that item in the queue finishes processing.

In general, I agree: it's quite hard to follow the logic. The only real solution I see is to go all in on an event-based API and drop the synchronous calls (which is the philosophy of https://github.com/TuyaAPI/driver).

tsightler:
Yeah, that's true if I just return the null response on timeout, which

tsightler:
OK, fixed the async issue by returning the promise from queue.add but also added an emit error event when there's a timeout by using p-timeout instead of depending on the queue timeout so I can run a function on timeout using the exact same approach as find().  Also fixed the package.json file.

Seems to work great for both sync/async, and timeouts return null, but also emit a timeout error.

index.js
```diff
@@ -128,6 +130,18 @@ class TuyaDevice extends EventEmitter {
       try {
         // Send request
         this._send(buffer).then(async data => {
```
codetheweb, line 132: Do we want to wrap get() in a queue too?

index.js
```diff
@@ -220,18 +234,28 @@ class TuyaDevice extends EventEmitter {
...
+            this._setResolver = resolve;
+          } catch (error) {
+            reject(error);
+          }
+        }), this._responseTimeout * 1000, () => {
+          // Only gets here on timeout so clear resolver function and emit error
+          this._setResolver = undefined;
+          this.emit('error', 'Timeout waiting for status response from device id: ' + this.device.id);
+        });
+        resolve(data);
```
codetheweb, line 254: I'm a little confused on how this works. Isn't `data` when `resolve()` called a Promise, not the result?

I think the top-level Promise (`this._setQueue.add(() => new Promise((resolve, reject) =>`) can be changed to an async arrow function to simplify some of this.

codetheweb (changes requested):

index.js
tsightler: I don't think there's any reason to wrap get() in a queue, as it doesn't suffer from the same issues as set() as it uses sequence numbers and adds the resolver to an array based on this sequence number.  The packet handler then resolves the correct send() based on the received sequence number so we can be assured that the response DP_QUERY returned by send() is properly associated to the request DP_QUERY and multiple outstanding requeues are no issue.  Also, send() should timeout after 5 retries to send the DP_QUERY, so I believe get() will properly timeout in most cases although there may still be a hang case where send() somehow to somehow send DP_QUERY, but no response DP_QUERY is received by the packet handler, so maybe an upper level timeout is a good idea.

However, with set(), a CONTROL command is used for send(), and indeed send() does received a CONTROL response with sequence number, but that CONTROL response is basically just an acknowledgement that the set happened so the code doesn't use the CONTROL response to resolve the set().  Instead it waits for the next STATUS message from the device and uses this to resolve the set().  The problem with this is that STATUS is not technically a response to CONTROL, the device just sends STATUS because something changed, so the STATUS command has no sequence number and thus there's no way to tie a specific STATUS to a specific CONTROL.

The code as it exist today just assumes that the next received STATUS is a response to the most recently sent CONTROL, which is reasonable, as long as there aren't multiple set() running concurrently or one of the sets doesn't get a response, which causes it to hang forever.  The only way to be reasonably sure that the next STATUS is a response to the CONTROL is to have only one set() at a time, this is especially true if you issue a CONTROL to a non-exist DPS, since the device will not acknowledge that with a STATUS message at all.

Now, with a little research, I may be able to figure out how to detect a CONTROL response to a non-exist DPS as I'm guessing the device probably returns an error in the CONTROL response.  Then we wouldn't have to rely on the timeout for that case. but I haven't gone down that path yet as my goal was just to try to make it work without hanging as it does now.

index.js
tsightler: This is most likely due to my limited Javascript abilities.  I feel like I've got a good handle on what the code is doing, but my skillset with Javascript is still quite basic as I'm just learning as I hack on these projects.

I originally did have an async arrow function there, but 'npm test' said this wasn't allowed (I don't remember the exact message but something about promise should not use async function?)

I originally tried to do this without the assignment to data at all, just return pTimeout (which is what happens in find) but somehow I could never get the syntax and behavior to work correctly.  I'll take another stab at it as I agree it's hard to follow.

tsightler: Updated with a new version that removes the extra, unnecessary promise.  Not sure why that was so hard for me the first time, seemed super clear today.  Let me know if you think this is better or could still use improvement.

codetheweb: Looks great!

codetheweb:
Thanks for all your hard work @tsightler!

Published in v6.0.1.

PR #380, opened 2020-11-28
Add option to disable initial get request

FliegenKLATSCH:
Firing an get here is bad style. If you just want to set an option, there is no need to get it.
Because it's a breaking change, I just added an option. Please consider changing this default behaviour in the next major release.


codetheweb:
I'm not sure it's "bad style", but does have potential to interfere with synchronous-style usage as you pointed out.

I updated the documentation and it looks good to me.

codetheweb (approved):

codetheweb:
Published in v6.1.0.

PR #381, opened 2020-11-28
Execute reject if not yet connected

FliegenKLATSCH:


FliegenKLATSCH:
@codetheweb It can still happen that neither resolve nor reject are called for some Promises (because there might be an error in the client), which means e.g. `await this.device.connect();` will never return / continue. Not sure what's the best approach to fix this.

codetheweb:
This check makes sense to me. Good to merge after those style issues are fixed (ignore the first errored build check, had to migrate to Travis CI .com).

What do you mean by
>  It can still happen that neither resolve nor reject are called for some Promises (because there might be an error in the client)

?

FliegenKLATSCH:
> those style issues are fixed
Which one? Maybe I am blind, but those checks don't say anything to me.

I mean, look at the change I did. I added a call of `reject()`. If the `this.client` throws an exception if the connection is already established (`this._connected == true`), then the `reject` is not called. It can still happen that none of the two (`reject`/`resolve`) is called, so the execution is stuck forever.

I ended up now using the "event approach" (ie. listening for the `error` event) instead of the async/promise callbacks. Maybe that should just be removed, although using it looks somehow more "modern", but I am not a JS specialist anyways.

codetheweb:
Linter run: https://travis-ci.com/github/codetheweb/tuyapi/jobs/451094811
You can run linter and tests locally with `npm test`.

> If the this.client throws an exception if the connection is already established (this._connected == true), then the reject is not called. It can still happen that none of the two (reject/resolve) is called, so the execution is stuck forever.

If I understand what you're saying, you think `connect()` should throw if errors happen after the initial connection? This wouldn't really make sense.

> Maybe that should just be removed, although using it looks somehow more "modern", but I am not a JS specialist anyways.

You're absolutely correct, IMO the event based approach works far better instead of the callback approach. However, removing support for callbacks would be a significant time commitment. You might have seen this before, but [this package](https://github.com/TuyaAPI/driver) was supposed to be the next version of TuyAPI (without callback support).

FliegenKLATSCH:
> Linter run: https://travis-ci.com/github/codetheweb/tuyapi/jobs/451094811
> You can run linter and tests locally with `npm test`.

I used the github web editor..

> > If the this.client throws an exception if the connection is already established (this._connected == true), then the reject is not called. It can still happen that none of the two (reject/resolve) is called, so the execution is stuck forever.
> 
> If I understand what you're saying, you think `connect()` should throw if errors happen after the initial connection? This wouldn't really make sense.

Maybe, but you got my point, that the flow is stuck, if neither resolve nor reject is called?

> > Maybe that should just be removed, although using it looks somehow more "modern", but I am not a JS specialist anyways.
> 
> You're absolutely correct, IMO the event based approach works far better instead of the callback approach. However, removing support for callbacks would be a significant time commitment. You might have seen this before, but [this package](https://github.com/TuyaAPI/driver) was supposed to be the next version of TuyAPI (without callback support).

Did not know about that other repository.

codetheweb:
> Maybe, but you got my point, that the flow is stuck, if neither resolve nor reject is called?

That won't happen; L327-338 will throw if a connection isn't established within the timeout period.

codetheweb:
Published in v6.1.1.

FliegenKLATSCH:
> > Maybe, but you got my point, that the flow is stuck, if neither resolve nor reject is called?
> 
> That won't happen; L327-338 will throw if a connection isn't established within the timeout period.

If a timeout happens, yes. Try it out and you may see other reasons. I tried with an unplugged plug ;)
... and as I said: it can still happen...

PR #405, opened 2021-01-12
Update SETUP.md to match new copyright

ThePheonixGuy:
Just followed this guide, and got lost on the Channel ID part.

Turns out on the developer platform, its now called URL Scheme in the modal when creating an app. 

It can also be seen later in step #3 of "Linking a Tuya device with Smart Link", in the model when linking your app.


codetheweb (approved):

codetheweb:
Looks good, thanks!

PR #406, opened 2021-01-13
Emit disconnected event when there's no heartbeat response

codetheweb:
Also adds a test to make sure it's possible to reconnect to a device after it goes offline.


codetheweb:
@ddv were you able to test this yet?

dvv:
@codetheweb Nope. I finally managed to flash them all with esphome.
Let us merge and see the feedback.
I'll publish the code I developed during tests before reflashing -- it may provide some ideas.

dvv:
As promised: https://github.com/dvv/tuya-mqtt/blob/master/tuya.js#L166-L221

codetheweb:
Published in v6.1.4.

Issue #412, opened 2021-01-29
not a bug!    ==>   Unable to get API-Key and details for TUYA


xX-Nexus-Xx:
I'm trying to use your TUYAPI modul but unable to get an API and other details from iot.tuya.com as it appears they have updated their site. It is now not possible to go into CLOUD Development as personal user. Therefore you are unable to create a project (to get the API keys) or to link any devices (for virualID)

Any idea on how this can be solved otherwise?

Thanks for your help

codetheweb:
Can you please post a screenshot of where you're getting stuck?

I have the "basic" user account plan and am still able to see cloud development.

jokecr:
I believe I see the same issue. It now shows as expired and the only tier available is $3000, yikes

![Capture](https://user-images.githubusercontent.com/5993614/106397091-91727480-63c0-11eb-84d7-d37b43733a97.PNG)
![Capture2](https://user-images.githubusercontent.com/5993614/106397224-18bfe800-63c1-11eb-8065-967c05f17043.PNG)

xX-Nexus-Xx:
Please see below. I as well had a chat with support team and they informed me that Cloud Development is no longer available for non-commercial users. Looks like they did some recent changes and you registered before

![image](https://user-images.githubusercontent.com/2195190/106400475-bc1d0580-6472-11eb-89b6-91c377bd271f.png)

This is the verification section
![image](https://user-images.githubusercontent.com/2195190/106400503-dd7df180-6472-11eb-9c6b-dee16625f802.png)

codetheweb:
:(

Tuya has done this before and then seemed to reverse the change, but maybe they're making it permanent now?

Not much I can do unfortunately. If Tuya continues to lock down their services, there's not really any alternative besides [flashing custom firmware](https://github.com/ct-Open-Source/tuya-convert).

Open to new ideas on how to link devices / get keys.

milo526:
And even the custom firmware is limited for new devices due to the use of newer chips and newer updates that come pre-installed on these devices :/

Apollon77:
The "old andoid app way" is still working ... lets hope at least a bit ...

alienator88:
The following steps work, I just created a new account today:

1. Open a Private/Incognito browser window
2. Navigate to iot.tuya.com
3. In the popup, set the Functional Cookies to No and Advertising Cookies to No
4. Submit Preferences and click Sign up to create a new account
5. Go to 10minutemail.com and grab a temporary email address
[...]

PR #422, opened 2021-02-14
Update SETUP.md

werkstrom:
As per https://github.com/codetheweb/tuyapi/issues/412


codetheweb:
It seems to still be working [as of yesterday](https://github.com/codetheweb/tuyapi/issues/412#issuecomment-778850532).

Please try again and let me know if it still isn't; then we can update the docs.

Issue #417, opened 2021-02-05
Data every second instead of 30s (command 18)


Dema323:
*Not a bug, more a question*

When my phone is off or the app is not open im getting new data every 30s from the power plug. (consumption data)
But when the i have Smart life open the data gets updated every 1-2s.
Has anyone been able to get new data every second or so without having the app open ?
Thanks

codetheweb:
This is because the devices are designed to have only one local connection at a time. From the README:

> 📝 Notes
> Only one TCP connection can be in use with a device at once. If using this, do not have the app on your phone open.

Dema323:
i think locally ? i used the example code from the github, also no api calls are counted on my tuya developer account

codetheweb:
Hmm, are you polling the device for updates or waiting for updates to be sent? Can you post your script?

tsightler:
Based on the research I've seen and a few dumps I've received from other users, when the app is open it's constantly sending a request to update the data, some mysterious "Command 18" type that seems un-documented.  Some additional details are here:

https://github.com/rospogrigio/localtuya/issues/194

codetheweb:
Huh, interesting.

I think in this case it's probably the user's responsibility to request updates with that if they want. @Dema323, would you be able to try adding that and see if it works? Might be worth adding a note to the README about.

tsightler:
@codetheweb How could the user request the updates if TuyAPI doesn't have any way to send a command 18 to the device?

From my perspective the work required is to first figure out the exact structure of the packet (it looks like a 3.1 protocol packet with command type 18 and otherwise empty, basically like a heartbeat packet).

I actually almost wonder if "command 18" is just a new type of heatbeat packet.  Perhaps a config option that uses command 18 instead of command 9 for heartbeat, just like some Tuya implementations have a config option to use command 13 instead of command 7 for control packets (I think TuyAPI also has an issue open about supporting devices that use command 13).

I had originally planned to try to do this work myself, but I've decided that Tuya seems intent on making it difficult to control your devices locally and I'm just going to give up my Tuya devices completely (I only have 4 left anyway, and 2 are easily converted to Tasmota, I'll just replace the other 2).  While I believe I could implement the required changes in TuyAPI, I'm just no longer sure it's worth the effort.

tsightler:
[...]

PR #423, opened 2021-02-15
Implementation of Command 18 ( DP refresh ) #417

stevoh6:
Will solve https://github.com/codetheweb/tuyapi/issues/417

@codetheweb I need your help, especially with nodejs. I implemented all required changes from tinytuya (this line should do the magic, together with new command definition https://github.com/jasonacox/tinytuya/blob/585827749653926dbbc372379a20b9c695d38d7b/tinytuya/__init__.py#L770), but after couple of hours, Im unable figured out, why Im getting this error, when I run `node _debug.js`

```
Connected to device!
DP_REFRESH data:  {
  payload: 'query dp failed',
  leftover: false,
  commandByte: 18,
  sequenceN: 1
}
Disconnected from device.
```


index.js
```diff
@@ -164,6 +166,75 @@ class TuyaDevice extends EventEmitter {
     });
   }
 
+  /**
+   * Gets a device's current status.
+   * Defaults to returning only the value of the first DPS index.
+   * @param {Object} [options]
+   * @param {Boolean} [options.schema]
+   * true to return entire list of properties from device
+   * @param {Number} [options.dps=1]
```
Apollon77, line 175: but options.dps can also be an array as I saw below, or?!

index.js
```diff
@@ -435,6 +509,12 @@ class TuyaDevice extends EventEmitter {
             await this._sendPing();
           }, this._pingPongPeriod * 1000);
 
+          // Automatically ask for dp_refresh so we
+          // can emit a `dp_refresh` event as soon as possible
+          if (this.globalOptions.issueRefreshOnConnect) {
+            this.refresh([4,5,6,18,19,20]);
```
Apollon77, line 515: this call is out of specs according to js doc and method abiove, or am I blind?

Apollon77: And please add a way to self specify a list of known dps if they are known to the application already

index.js
```diff
@@ -38,21 +38,23 @@ const {UDP_KEY} = require('./lib/config');
...
+                issueGetOnConnect = true,
+                issueRefreshOnConnect = false
+              } = {}) {
     super();
     // Set device to user-passed options
     this.device = {ip, port, id, gwID, key, productKey, version};
     this.globalOptions = {
-      issueGetOnConnect
+      issueGetOnConnect,
+      issueRefreshOnConnect
```
Apollon77, line 57: do we need to add checks if someone provides both? or should refresh noch be "the fallback" if get do not work?

Apollon77 (commented):
SOme comments from me :-)

stevoh6:
Hi  @Apollon77 thanks valid points, but I was trying focus on prove of concept, after that I clean it up. Now its just testing branch :D 

Do you know why Im not getting correct data? What could be potential problem? I should get in response changed dps, e.g. {19:80} if dps 19 was changed. Thx.

index.js
stevoh6: yes you are sendig dps, which you would like to "monitor" ( get update from them)

index.js
stevoh6: make sense, no idea yet, how to do it :D

stevoh6:
Fixed,  Im able to retrieve real-time data from device. So I can put it together and make some cleanup :-)
This is correct payload ( especially we need to use `dpId` instead of `dps`
```
  TuyAPI IP and ID are already both resolved. +0ms
  TuyAPI Connecting to 192.168.***.***... +4ms
  TuyAPI Socket connected. +8ms
Connected to device!
  TuyAPI GET Payload: +2ms
  TuyAPI {
  TuyAPI   gwId: 'bf73f75cfc439********',
  TuyAPI   devId: 'bf73f75cfc4399ea*******',
  TuyAPI   t: '1613382457',
  TuyAPI   dpId: [ 18, 19, 20 ],
  TuyAPI   uid: 'bf73f75cfc4399ea*******'
  TuyAPI } +1ms
  TuyAPI Received data: ********** +22ms
  TuyAPI Parsed: +1ms
  TuyAPI {
  TuyAPI   payload: { dps: { '18': 89, '20': 2430 }, t: 1613382513 },
  TuyAPI   leftover: false,
  TuyAPI   commandByte: 8,
  TuyAPI   sequenceN: 0
  TuyAPI } +0ms
````

**Comment by stevoh6** [comment]
Now seems to be working pretty well. Please check/test and provide feedback.

```
>node _debug.js
Connected to device!
DP_REFRESH data:  { dps: { '19': 127 }, t: 1613388138 }
Data from device: {
  dps: {
    '1': true,
    '9': 0,
    '17': 10,
    '18': 87,
    '19': 127,
    '20': 2466,
    '38': 'off',
    '41': '',
    '42': '',
    '43': ''
  }
}
Boolean status of default property: true.
Data from device: { dps: { '1': false }, t: 1613388139 }
Boolean status of default property: false.
Disconnected from device.
```

File: index.js [hunk 4]
```diff
@@ -160,7 +169,76 @@ class TuyaDevice extends EventEmitter {
...
+   * tuya.refresh({schema: true}).then(data => console.log(data))
+   * @returns {Promise<Boolean|Object>}
+   * returns boolean if single property is requested, otherwise returns object of results
+   */
+  refresh(options = {}) {
+    const payload = {
+      gwId: this.device.gwID,
+      devId: this.device.id,
+      t: Math.round(new Date().getTime() / 1000).toString(),
+      dpId: this._dpRefreshIds,
```
codetheweb, line 201: This is consuming the instance options instead of the function options, if that's the intended behavior the above documentation needs to be updated.

codetheweb (changes requested):
Looks pretty good @stevoh6, thanks for taking the time to contribute.

index.js
stevoh6: Nope, it was mistake. Now it should be okay.

stevoh6:
> Looks pretty good @stevoh6, thanks for taking the time to contribute.

Thx, I hope that it helps lots of projects to move forward.


P.S. solution was found by project  https://github.com/jasonacox/tinytuya, all glory to them, I just helped test and replicate it. :-)

codetheweb:
LGTM. One last thing; could you add a note to the README about when refresh is needed (or when you should pass it as an option in the constructor)?

Apollon77:
Honestly I would have loved to see it like in that referenced pythion variant ... "Implicitely done when get returns data unavailable/unvalid json error". And if happened once just remembering it for this instance. Now I need to implement that outside of the lib - or as "dumb dev" i simply set it always ... and also to hand over the known list of dps in constructor to use that for the initial refresh instead of the hard coded value list would also be more convenient for usages that "Know" the schema

Apollon77:
@stevoh6 What do you think about my comments?

codetheweb:
Bump. Ready to merge as soon as https://github.com/codetheweb/tuyapi/pull/423#issuecomment-780929733 is resolved.

@Apollon77 if I'm understanding what you're saying corectly I think that could be addressed in a seperate PR.

Apollon77:
yes could :-) Soif you merge maybe wait  short before releaeing and I provide a PR as I think it would make sense. 
But also incuding here is an option if @stevoh6 minds

stevoh6:
Hi, first off all Im sorry, I didnt have enough time to finish it. 

@Apollon77 definitely great suggestion, but I will personally do it in separate-following PR, to not block other projects ( as @codetheweb  suggest).

codetheweb:
No problem, looks good.

One last thing that I should have thought about earlier: is there a reason not to combine `dp-refresh` and `data` events? (They can still be distinguished by the command byte if needed.)

stevoh6:
Its comming from my prax, dont break existing functionality, dont call something twice, if you need one call and lets user/programmer deside what is the best scenario for him. In this case you dont want to receive alway status of switch every 10sec, you want to only get power data. Alteast for it make sense in mqtt scenario.

codetheweb (approved):

codetheweb:
Published in v6.2.0.

PR #439, opened 2021-03-10
Update iot.tuya.com instructions for new interface

alexlmiller:
The interface for iot.tuya.com change and no longer has the "API Groups" interface but rather now "API Products" with a different mechanism for enabling them.  I've compiled the instructions and validated they work as of 3/10/2021


codetheweb:
THank you!

PR #465, opened 2021-05-08
Add ability to get and set zigbee gateway devices by specifying additional attribute cid

imbenwolf:


index.js
```diff
@@ -146,7 +147,7 @@ class TuyaDevice extends EventEmitter {
     return new Promise((resolve, reject) => {
       // Send request
       this._send(buffer).then(async data => {
-        if (data === 'json obj data unvalid' && options.schema !== true) {
+        if (data === 'json obj data unvalid' && options.schema === true) {
```
Apollon77, line 150: Why that change?

imbenwolf: honestly, the only reason is when i issue a normal get request to a non-zigbee device through https://github.com/TuyaAPI/cli/blob/master/lib/control.js#L49 (where schema is set to true) it doesn't send a null set request to get the status and it just resolves to "json obj data unvalid". i didn't really understand what the schema option should be doing reading through the code...

index.js
```diff
@@ -129,6 +129,7 @@ class TuyaDevice extends EventEmitter {
       devId: this.device.id,
       t: Math.round(new Date().getTime() / 1000).toString(),
       dps: {},
+      cid: options.cid,
```
codetheweb, line 132: This needs to get added to the above documentation.

index.js
codetheweb: The `schema` option is a result of poor API design on my part 😛. It controls the data that's passed back through TuyAPI, not the contents of the payload itself.

I'm not entirely sure why we're checking the schema option here; I think that condition should just be dropped.

index.js
```diff
@@ -306,14 +307,22 @@ class TuyaDevice extends EventEmitter {
...
-      devId: options.devId || this.device.id,
-      gwId: this.device.gwID,
-      uid: '',
+    let payload = {
       t: timeStamp,
       dps
     };
 
+    if (options.cid) {
+      payload.cid = options.cid;
```
codetheweb, line 316: Again, docs need to be updated for this parameter.

index.js
```diff
@@ -339,7 +348,7 @@ class TuyaDevice extends EventEmitter {
       } catch (error) {
         reject(error);
       }
-    }), this._responseTimeout * 1000, () => {
+    }), this._responseTimeout * 5000, () => {
```
codetheweb, line 351: Rational on increasing the timeout? I'm ok with bumping it a bit but 5s seems excessive.

codetheweb (changes requested):
I am curious why you'd use a Tuya zigbee hub rather than one that's more friendly to work with but this looks pretty good to me.

index.js
Apollon77: Honestly I also struggled on last PR and "usage" on that one and implemented it by myself outside... so it is a change in behavour if we kill the flag (and I'm up for it!!) but should be clearly stated (but I assume not breaking)

codetheweb: Yeah, I don't think removing the check for `schema` here should be breaking. I'm thinking `schema` is still used to control the shape of return data for backwards compatibility.

imbenwolf: I agree, schema as a flag to send all attributes is good for the moment (we can think about changes in the future, like it doesn't fully make sense to me that the default behaviour is to return only all the dps instead of the whole response -> if i send a request, i want the whole response, not just a part). But I agree that removing the check is necessary. I removed it in the latest commit

index.js
imbenwolf: Done! Although I now changed cid attribute to be a tuya device attribute instead of a get/set function attribute in the latest commit :)

index.js
imbenwolf: Done!

index.js
imbenwolf: You're right, I tested it out with 2.5s and it works fine on my side. 1s was just not enough time to get the DP_REFRESH after a SET request and I agree that 5s is excessive. I reduced it in my latest commit :)

index.js
codetheweb: I could be wrong on this but would it make more sense to have it passed in the function parameters like you had before? Don't you need multiple instances of TuyAPI now to talk to a single gateway if there's multiple subdevices connected to it?

imbenwolf: I guess it depends how we want to look at it from an API perspective: 
- 1 device (gateway) controlling multiple subdevices by cid
- or multiple devices (e.g lock, heating) controlled through the same ip & id

what do you prefer?

codetheweb: I would say the second, but thinking through this some more I don't think this would work in practice since Tuya devices tend to be flaky if more than one TCP connection is open to them at a time. So let's go with the first for now (include `cid` in get / set options rather than constructor).

imbenwolf: Done! :)

codetheweb:
Thanks for your work!

codetheweb (approved):

codetheweb:
Published in v7.2.0.

PR #581, opened 2022-06-10
Some devices only send information on request (Command 18)

markuslaube:
Some devices only send information on request (Command 18).
This update allows using the variable (issueRefreshOnPing = true) to send the command 18 after each ping.


markuslaube:
uncomplete

PR #582, opened 2022-06-10
Command18 after ping via Option

markuslaube:
This update allows using the variable (issueRefreshOnPing = true) to send command 18 after each ping and then to query the status using get. Since there are now some devices that do not send updates by themselves.


codetheweb (changes requested):
Looks good! Just need to add a note to the JSDoc too.

markuslaube:
ihm not a developer I hope that's it this what's you mean ;)

Apollon77:
One dumb question ... DIO we need the Ping at all when we send a real message? ;-)

markuslaube:
> One dumb question ... DIO we need the Ping at all when we send a real message? ;-)

I have no idea :-O

Apollon77:
... just question because the "Ping" is mainly there to make xure the connection stays open when I remember correctly ... wehen we now send other data it should have the same purpose ... but ok we also do not need to experiment that out ... ;-)

markuslaube:
Hello @codetheweb ,

please can you say if the change on [docs/index.html]this what's you mean with "Just need to add a note to the JSDoc too." or is this a misunderstanding from me?

Thanks, Markus

codetheweb (approved):
Yep, thanks.

codetheweb:
Released in v7.3.0.

Issue #320, opened 2020-05-11
Add TypeScript type definitions 


grenewode:
Currently, the project does not provide any type definitions, which makes it difficult to use in typescript projects. 

Since the project is already extensively commented with jsdoc comments, much of the work is already done. There are still places where updates will need to be made, but those should be largely trivial, just a matter of adding more jsdoc comments to provide type hints where typescript cannot infer types.

I'd be happy to work on this.

codetheweb:
I'd love to have type definitions, go for it 👍.

hwr7dd:
This would be super helpful if someone is able to work on this

duncte123:
I would love to take on this issue if it's still available

codetheweb:
Go for it 🙂

PR #592, opened 2022-07-19
Add typedefs

duncte123:
Closes https://github.com/codetheweb/tuyapi/issues/320

I tested this with `npm link` to make sure that the types are getting picked up correctly in my IDE

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,86 @@
+declare module 'tuyapi' {
+    import { EventEmitter } from 'events';
+
+    interface TuyaDeviceOptions {
+        ip?: string;
+        port?: number;
+        id: string;
+        gwID?: string;
+        key: string;
+        productKey?: string;
+        version?: number|string;
+        nullPayloadOnJSONError?: boolean;
+        issueGetOnConnect?: boolean;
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -0,0 +1,86 @@
...
+        version?: number|string;
+        nullPayloadOnJSONError?: boolean;
+        issueGetOnConnect?: boolean;
+        issueRefreshOnConnect?: boolean;
+        issueRefreshOnPing?: boolean;
+    }
+
+    type UnionTypes = Object|number|string|boolean;
+
+    interface Object {
```
codetheweb, line 20: This could probably just be `Record<string, unknown>`

codetheweb (commented):
Thank you!

index.d.ts
duncte123: I think the `any` type might be a better fit in this case

codetheweb: I would rather have `unknown` to force consumers to type the expected payload themselves (since it varies so much from device-to-device) but `any` is fine for now. 😃

codetheweb (approved):

codetheweb:
Published in v7.4.0.

PR #606, opened 2022-11-06
Protocol 3.4 support

### LINKED ISSUE #234: Add support for v3.2 protocol
Bonjour,

J'obtiens cette erreur au demarrage de gladys:

` TypeError: Cannot read property '1' of undefined
0|gladys   |     at _send.then.data (/home/pi/gladys/api/hooks/tuya/node_modules/tuyapi/index.js:123:41)
0|gladys   |     at <anonymous>
0|gladys   |     at process._tickDomainCallback (internal/process/next_tick.js:228:7)`

et des que j'actionne le device:

`0|gladys   | (node:18902) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): TypeError: Cannot read property '1' of undefined
0|gladys   | Tuya - Error, undefined device!
0|gladys   | Sending 500 ("Server Error") response:
0|gladys   |  Error: find() timed out. Is the device powered on and the ID or IP correct?
0|gladys   |     at pTimeout (/home/pi/gladys/api/hooks/tuya/node_modules/tuyapi/index.js:593:19)
0|gladys   |     at Timeout.setTimeout [as _onTimeout] (/home/pi/gladys/api/hooks/tuya/node_modules/p-timeout/index.js:20:13)
0|gladys   |     at ontimeout (timers.js:475:11)
0|gladys   |     at tryOnTimeout (timers.js:310:5)
0|gladys   |     at Timer.listOnTimeout (timers.js:270:5)
0|gladys   | Server Error:
0|gladys   |`

ligne 123 de tuyapi/index.js:
` resolve(data.dps['1']);`

J'utilise un smart switch:
SCW NF101

avec wireshark, je recupere:
 `ip":"192.168.1.xx","gwId":"bfx9268f215ecx78cdkpiv","active":2,"ablilty":0,"encrypt":true,"productKey":"zWxOK04ovOjw0Pxw","version":"3.2"}.B0`

Mes identifinats sont de cette forme:
`[ { id: 'bfx9268f215ecx78cdkpiv', key: 'b6a93451d8f07ga2' } ]`

des idees?

### LINKED ISSUE #481: Support protocol version 3.4
**Describe the bug**
I tried the example script from the read me. Tuyapi is able to find the device and to connect, but directly disconnects afterwards. It's not possible to keep the connection and listen to data. 

Also tried it in Node-Red with package node-red-contrib-tuya-smart-device which uses latest tuyapi. Same error, it constantly reconnects. But at least it receives some data events which results in error logs with messages like `Error: CRC mismatch: expected -674619356, was -22843090`

Maybe the problem is the protocol version 3.4? Didn't find any information if tuyapi supports 3.4.

debug output:
```
node test.js
  TuyAPI Finding missing IP undefined or ID bf97ae127518bd821b1mdo +0ms
  TuyAPI Received UDP message. +4s
  TuyAPI UDP data: +1ms
  TuyAPI {
  TuyAPI   payload: {
  TuyAPI     ip: '192.168.0.42',
  TuyAPI     gwId: 'bf97ae127518bd821b1mdo',
  TuyAPI     active: 2,
  TuyAPI     ablilty: 0,
  TuyAPI     encrypt: true,
  TuyAPI     productKey: 'keyyj3fy8x98arty',
  TuyAPI     version: '3.4',
  TuyAPI     token: true
  TuyAPI   },
  TuyAPI   leftover: false,
  TuyAPI   commandByte: 35,
  TuyAPI   sequenceN: 0
  TuyAPI } +1ms
  TuyAPI Connecting to 192.168.0.42... +5ms
  TuyAPI Socket connected. +7ms
Connected to device!
  TuyAPI GET Payload: +2ms
  TuyAPI {
  TuyAPI   gwId: 'bf97ae127518bd821b1mdo',
  TuyAPI   devId: 'bf97ae127518bd821b1mdo',
  TuyAPI   t: '1623446698',
  TuyAPI   dpId: [ 18, 19, 20 ],
  TuyAPI   uid: 'bf97ae127518bd821b1mdo'
  TuyAPI } +0ms
  TuyAPI GET Payload: +3ms
  TuyAPI {
  TuyAPI   gwId: 'bf97ae127518bd821b1mdo',
  TuyAPI   devId: 'bf97ae127518bd821b1mdo',
  TuyAPI   t: '1623446698',
  TuyAPI   dps: {},
  TuyAPI   uid: 'bf97ae127518bd821b1mdo'
  TuyAPI } +0ms
  TuyAPI Socket closed: 192.168.0.42 +13ms
  TuyAPI Disconnect +1ms
Disconnected from device.
```

Apollon77:
I catched the code from @harryzz repoditory where he posted that protocol 3.4 was implemented (see https://github.com/codetheweb/tuyapi/issues/481#issuecomment-985988646). I adjusted code style to the project one and reworked the code a bit to be less clustered when it comes to the 3.4 changes.
Additionally I added  a deferred promise for the "longer connect flow" to resolve the connect promise when the connection is really established.

This is completely untested! I hope that I get such a device tomorrow/tuesday, but if someone wants to : install from my Fork (https://github.com/Apollon77/tuyapi) and give it a shot. Ideally start with DEBUG=TuyAPI* and post your log when you tested it (successfuly or not) :-)

@codetheweb I will do a Draft PR for now, but if you have time please already have a look and tell me if you would like somethings to be adjusted.
I do not get any lint or tests running locally ... will try again, but if you have ideas....

fixes #234, fixes #481

### DIFF
--- a/index.js
+++ b/index.js
@@ -30,7 +30,7 @@ const {UDP_KEY} = require('./lib/config');
  * @param {Number} [options.version=3.1] protocol version
  * @param {Boolean} [options.nullPayloadOnJSONError=false] if true, emits a data event
  * containing a payload of null values for on-device JSON parsing errors
- * @param {Boolean} [options.issueGetOnConnect=false] if true, sends GET request after
+ * @param {Boolean} [options.issueGetOnConnect=true] if true, sends GET request after
  * connection is established. This should probably be `false` in synchronous usage.
  * @param {Boolean} [options.issueRefreshOnConnect=false] if true, sends DP_REFRESH request after
  * connection is established. This should probably be `false` in synchronous usage.
@@ -784,7 +784,7 @@ class TuyaDevice extends EventEmitter {
 
     // Status response to SET command
 
-    // 3.4 response sequenceN is not '0' just next
+    // 3.4 response sequenceN is not '0' just next TODO verify
     if (/*packet.sequenceN === 0 &&*/
         packet.commandByte === CommandType.STATUS &&
         typeof this._setResolver === 'function') {
--- a/index.js
+++ b/index.js
@@ -19,7 +19,7 @@ const {UDP_KEY} = require('./lib/config');
  * you're experiencing problems when only passing
  * one, try passing both if possible.
  * @class
- * @param {Object} options
+ * @param {Object} options Options object
  * @param {String} [options.ip] IP of device
  * @param {Number} [options.port=6668] port of device
  * @param {String} [options.id] ID of device (also called `devId`)
@@ -108,13 +108,13 @@ class TuyaDevice extends EventEmitter {
     this._dpRefreshIds = [4, 5, 6, 18, 19, 20];
     this._tmpLocalKey = null;
     this._tmpRemoteKey = null;
-    this.session_key = null;
+    this.sessionKey = null;
   }
 
   /**
    * Gets a device's current status.
    * Defaults to returning only the value of the first DPS index.
-   * @param {Object} [options]
+   * @param {Object} [options] Options object
    * @param {Boolean} [options.schema]
    * true to return entire list of properties from device
    * @param {Number} [options.dps=1]
@@ -162,8 +162,7 @@ class TuyaDevice extends EventEmitter {
     return new Promise((resolve, reject) => {
       // Send request
       this._send(buffer).then(async data => {
-        //console.log("DATA:" + data)
-        if (data === 'json obj data unvalid' /*|| data === 'data format error' || data === 'devid not found'*/) {
+        if (data === 'json obj data unvalid' /* || data === 'data format error' || data === 'devid not found' */) {
           // Some devices don't respond to DP_QUERY so, for DPS get commands, fall
           // back to using SEND with null value. This appears to always work as
           // long as the DPS key exist on the device.
@@ -193,7 +192,7 @@ class TuyaDevice extends EventEmitter {
   /**
    * Refresh a device's current status.
    * Defaults to returning all values.
-   * @param {Object} [options]
+   * @param {Object} [options] Options object
    * @param {Boolean} [options.schema]
    * true to return entire list of properties from device
    * @param {Number} [options.dps=1]
@@ -270,7 +269,7 @@ class TuyaDevice extends EventEmitter {
 
   /**
    * Sets a property on a device.
-   * @param {Object} options
+   * @param {Object} options Options object
    * @param {Number} [options.dps=1] DPS index to set
    * @param {*} [options.set] value to set
    * @param {String} [options.cid]
@@ -346,6 +345,7 @@ class TuyaDevice extends EventEmitter {
         ...payload
       };
     }
+
     if (this.device.version === '3.4') {
       /*
       {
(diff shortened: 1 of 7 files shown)


Apollon77:
@codetheweb Ok, lint and tests work now ... (and I adjusted a lot lint stuff)

codetheweb:
Thanks for doing this! Just wanted to let you know I've been really busy the last few days but hope to take a look soon.

Apollon77:
@codetheweb In the meantime I should also have a 3.4 device ... So plan is to try the changes out myself today

index.js
```diff
@@ -439,6 +474,62 @@ class TuyaDevice extends EventEmitter {
     }
   }
 
+  /**
+   * Create a deferred promise that resolves as soon as the connection is established.
+   */
+  createDeferredConnectPromise() {
```
codetheweb, line 480: I don't think we need this, can we just return `new Promise()` at the end?

codetheweb (commented):
Trusting that this all works since I don't have a device. :)

Tests pass locally for me 👍, not sure why they're not running on CI.

index.js
Apollon77: The problem is that the 3.4 "connect" is in the end 3 messages exchanged, sothe final "resolve" will come from out an other method but we also still want to have "connect timeout" logic. So in the end for a "clean" solution we in any case need to store a promise.

The alternative is that we resolve connect promise already after we sent out the "start session exchange message", but in fact we are not yet connected.

Apollon77:
For testing: I have one device and tested with this,but in the end no" long time testing" so far. I have also one more user using it - he had one time an issue that the devices went into a "restart loop", but honestly ... This could mean something is not 100%, but considering the quelity sometimes it must not mean anything ;-) And also after restarting problem did not re-occured till now.

So I would consider as ok aka "best we can say for now" :-)

Apollon77:
@codetheweb Lets wait a day longer ... On user that tried this has oissues, so I now started a bit a "Long time test" ... lets see if it is stable for me

Apollon77:
@codetheweb Ok I would go and release this. In fact it works. I only have one device to verify and yes this device closes the connection every 2-3h, but a reconnect works after 30 or latest second try after 60s ... but from all the quality issues we know from Tuya devices this is not enough to proof it a "protocol implementation issue" (and also python lib do not have anything else). So best it can be right now

index.js
codetheweb: Right, I agree we need to store a Promise, just not sure why we need to have this wrapper that doesn't seem to have any custom logic to create one.

Apollon77: This wrapper is needed because you can not just store the two methods reject and resolve... I experimented around a long time on how to build a deferred promise and this solution is used in one of my other projects and originally came from an other developer that uses it in his projects.

In fact if you return a promise you have no real way to reference to the reject or resolve method of this object "outside" of the Promise constructor. Thats why this is the trick to allow this.

It cost me hours to come to this in my other project :-)

You find it also here e.g. https://www.getrevue.co/profile/masteringjs/issues/promises-and-the-deferred-antipattern-235313

Apollon77: (and I decided against adding a dependency lib to get a Deffered option which is kind of just doing exactly also that - e.g. https://github.com/ljharb/promise-deferred/blob/main/index.js

Apollon77: PS: If you mean that I should kill the extra function and directly integrate the code into the connect function then yes I can do this. ;-)

codetheweb: Read the code again and I understand the issue now, seems ok for now.

codetheweb (approved):

codetheweb:
Published in v7.5.0.

Apollon77:
Thank you very much!

PR #610, opened 2022-11-14
fix discovery UDP package handling for protocol 3.4

Apollon77:
@codetheweb SOrry, we need to publish a fix ... sorry ... There was somethign wrong with UDP discovery package handling for 3.4 protocol that I didn't noticed before


codetheweb (approved):

codetheweb:
Released in v7.5.1.

PR #616, opened 2022-12-05
add special handling for SET-as-GET requests that return DP_QUERY

### LINKED ISSUE #617: get() timeout error after the result was successfully retrieved
### Discussed in https://github.com/codetheweb/tuyapi/discussions/613

<div type='discussions-op-text'>

<sup>Originally posted by **kardoka** November 21, 2022</sup>
Greetings, I am getting a timeout error for the `get()` command after I successfully process the result of it. What could be the issue here? Would it be possible to handle this error?

The error message (note that I am hiding the device ids and full path to working directory):
```
node:internal/errors:478
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_UNHANDLED_ERROR]: Unhandled error. ('Timeout waiting for status response from device id: hidden)
    at new NodeError (node:internal/errors:387:5)
    at TuyaDevice.emit (node:events:502:17)
    at ./node_modules/tuyapi/index.js:411:12
    at Timeout._onTimeout (./node_modules/p-timeout/index.js:25:13)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7) {
  code: 'ERR_UNHANDLED_ERROR',
  context: 'Timeout waiting for status response from device id:  hidden
}
```

The code itself:
```
const TuyAPI = require('tuyapi');
require('dotenv').config();

const device = new TuyAPI({
  id: process.env.GATEWAY_DEVICE_ID,
  key: process.env.LOCAL_KEY,
  version: 3.3,
  issueRefreshOnConnect: true,
});

(async () => {
  await device.find();

  await device.connect();

  let status = await device.get({
    cid: process.env.CHILD_ID,
    dps: 1
  });

  console.log(`Current status: ${status}.`);

  device.disconnect();
})();
```

EDIT: I am able to catch it using:

```
process.on('uncaughtException', error => {
  console.log('Caught it!');
})
```

But is there some more elegant way to catch it or to avoid this exception altogether?</div>

Apollon77:
fixes #617 
fixes https://github.com/codetheweb/tuyapi/discussions/613

This is untested currently!

### DIFF
--- a/index.js
+++ b/index.js
@@ -774,7 +774,17 @@ class TuyaDevice extends EventEmitter {
 
     // Returned DP refresh response is always empty. Device respond with command 8 without dps 1 instead.
     if (packet.commandByte === CommandType.DP_REFRESH) {
-      debug('Received DP_REFRESH empty response packet.');
+      // If we did not got any STATUS packet, we need to resolve the promise.
+      if (typeof this._setResolver === 'function') {
+        debug('Received DP_REFRESH empty response packet without STATUS packet');
+        this._setResolver(packet.payload);
+
+        // Remove resolver
+        this._setResolver = undefined;
+        this._setResolveAllowGet = undefined;
+      } else {
+        debug('Received DP_REFRESH response packet.');
+      }
       return;
     }
 
--- a/index.js
+++ b/index.js
@@ -234,7 +234,7 @@ class TuyaDevice extends EventEmitter {
       payload.cid = options.cid;
     }
 
-    debug('GET Payload:');
+    debug('GET Payload (refresh):');
     debug(payload);
 
     // Create byte buffer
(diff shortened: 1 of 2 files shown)


Apollon77:
@codetheweb I asked the user that is affected by the issue to check my branch. I tried to make it backward compatible and stumbled over the case that two parallel "set" in general would be an issue because in this case the setResolve variable is overwritten. I added a check and throw an error in that case. This is slightly breaking, but I think the much cleaner way. What do you think?

codetheweb:
Yeah I think that makes sense.

Will merge once we get confirmation from the user.

Apollon77:
@codetheweb I added another case that I stumbled over with a user (the device was just sending the REFRESH response, but did not sent any STATUS package in between, so we never resolved the promise. The usual process is

-> send DP_REFESH
-> receive STATUS with data
-> receive "empty" DP_REFRESH

But it seems that it also can happen like:

-> send DP_REFESH
-> receive "empty" DP_REFRESH

without data. So I now check if the resolver is still there and resolve on DP_REFRESH too with the empty package.

A useris testing this right now...

Apollon77:
@codetheweb Verified ... it seems we never really resolved for "refresh" (was only working when falling back to "set") ... we just emitted the data ...  veeery interesting

codetheweb (approved):
nice, thanks for the good work :)

codetheweb:
Published in v7.5.2.

Also added you as a contributor to this repo

Apollon77:
Thank you!

PR #651, opened 2024-06-23
Create SECURITY.md

karnelius-hub:


Apollon77:
And why we should use a dummy SECURITY.md file? There is no actual meaningful content in it

PR #659, opened 2024-08-16
increase responseTimoeut to 5 seconds

jonnytest1:
thats literally it in the title 

i had some commands failing but overwrtiing the repsonse timeout to a bit longer helped

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,7 +92,7 @@ class TuyaDevice extends EventEmitter {
     // Socket connected state
     this._connected = false;
 
-    this._responseTimeout = 2; // Seconds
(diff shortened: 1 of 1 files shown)


Apollon77:
Hm ... the default timeout seems to work just fine .. so I would more think it could be an idea to allow override by constructor option or - if it is "command specific" then also an override for the "Just send" method ... WDYT?

Apollon77:
@jonnytest1 do you want to provide an update to our pr or should I do when back from vacation in a week?

Apollon77:
No response, so I close here

PR #690, opened 2025-01-27
Optimize 3.5 detection and prevent crashes

Apollon77:
see commits


nospam2k:
I've switched to using Home Assistant so I'm not actively using Tuyapi, Sorry.

Issue #691, opened 2025-01-29
Open to a rewrite in Typescript?


ClementNerma:
Hi there!

Native Typescript code (with builtin types instead of a separate declaration file, which sometimes causes problems depending in TS projects depending on their config) is IMO preferable to Javascript in most cases.

Would you be open to a PR rewriting the project's code in TypeScript? If so, I'd be happy to make one.

Otherwise, no problem!

Apollon77:
I would be more then open for that ... I also have that on my list but not yet time to start with it. most of my other maintained projects currently are TS so it is always a pain to do pure JS :-)

I will finish current PR and then would be ideal time tbh

Apollon77:
@codetheweb any veto?

codetheweb:
I planned to many years ago. Sounds good to me.

ClementNerma:
I've started rewriting the project in TypeScript, but I'm a bit stuck as there are tons of conflicting types in various places.
For instance, take the `_decryptPre34` function:

![Image](https://github.com/user-attachments/assets/1755c03e-6474-4ca6-a02b-79fec36be511)

No matter how I type it, it can't work. The `decipher.update` method takes a `string`, followed by an `Encoding`, followed by an `OutputEncoding`.

The thing is, `Encoding` can never be `"buffer"` as per Node.js' documentation. Also, the `data` cannot be a `Buffer`, it must be a `string`. So I don't know what's supposed to happen here.

There are lots of other things like that:

![Image](https://github.com/user-attachments/assets/d29dbc11-e338-448e-9c90-ba5c393a7b44)

The `getPayload` method may return `false` if for instance the provided string or buffer is empty. But in `Packet`, the `payload` property cannot be `false`.

Also, the docs show that its type should be `Buffer | Object | string`, but when we provide it to `getPayload` it can't work as this method then provides it to `this.cipher.decrypt`, which only takes a `Buffer`, not an `Object` or a `string`. So I'm a bit confused...

Could you please enlighten me on these examples? Thanks.

Apollon77:
@ClementNerma these are the interesting pars here... I also have not that many examples.

Following logs in other issues here the data received on UDP socket level should be Buffer ... thats means you are right "prsePayload" and the package definition from they is always a Buffer for the payload. We could call this the "Internal Payload" i guess ... after that the payload is tried to be parsed into  more convenient variant which might be object/string/boolean ... 

[...]

PR #692, opened 2025-02-09
Partial rewrite in TypeScript

ClementNerma:
Rewrite of TuyAPI in TypeScript.

There are lots of things in the code I don't understand so I refrained from rewriting them to avoid introducing new issues. So this is only a partial rewrite, most of the code has been rewritten, except for `index.js` which still has lot of unrtyped code.

From #691


package.json
```diff
@@ -1,73 +1,56 @@
...
+	},
+	"homepage": "https://github.com/codetheweb/tuyapi#readme",
+	"dependencies": {
+		"debug": "^4.3.7",
+		"p-queue": "6.6.2",
+		"p-retry": "4.6.2",
+		"p-timeout": "3.2.0"
+	},
+	"devDependencies": {
+		"@biomejs/biome": "^1.9.4",
```
Apollon77, line 45: I personallymore tend to stay with eslint and prettier than using "just another package that somehow contains simmilar things" ...

Apollon77 (changes requested):
Sorry for the late review, but here we go. Thats a cool start. I have some smaller comments - and sorry that you need to rebase because of some fixes I did :-(

Formatting wise please stay with spaces instead of tabs .. I am fully ok with 4 spaces, but please no tabs :-)

ClementNerma:
Hi, I just realized this PR was still open ; I don't have time to take care of it currently so I'm gonna go ahead and close it. If I have time in the future I'll reopen one :)
Have a nice day!

Apollon77:
Ok, maybe I give it a shot then somewhen

